PR #25, opened 2019-06-26
Shut down ML queue when app enters background

kingst:


jaimejiyepark:
Looks great! #shipit.

PR #26, opened 2019-06-27
bundle moved to model

jaimejiyepark:


BundleURL.swift
```diff
@@ -0,0 +1,20 @@
+import CoreML
+
+@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
+class var BundleURL: URL{
+    /// URL of model assuming it was installed in the same bundle as this class
+    let documentDirectory = try! FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor:nil, create:false)
```
kingst: Still a !

BundleURL.swift
```diff
@@ -0,0 +1,20 @@
+import CoreML
+
+@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
+class var BundleURL: URL{
```
kingst: Maybe make this a struct and have a static method called `bundleUrl(modelName: String) -> URL?` or something like that

BundleURL.swift
```diff
@@ -0,0 +1,20 @@
...
+    /// URL of model assuming it was installed in the same bundle as this class
+    let documentDirectory = try! FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor:nil, create:false)
+    let modelcFile = documentDirectory.appendingPathComponent("FindFour.mlmodelc")
+    
+    if !FileManager.default.fileExists(atPath: modelcFile.path) {
+        if let bundleUrl = Bundle(for: FindFour.self).url(forResource: "CardScan", withExtension: "bundle"){
```
kingst: I suggest using

```
guard if let bundleUrl = Bundle(for: ...) else {
    print("could not get bundle url")
    return nil
}

guard if let bundle = Bundle(url: bundleUrl) else {
    return nil
}
```

and so on

BundleURL.swift
```diff
@@ -0,0 +1,20 @@
...
+            if let bundle = Bundle(url: bundleUrl){
+                if let modelUrl = bundle.url(forResource: "FindFour", withExtension: "bin"){
```
kingst: Same thing here with guard, we should try to avoid the if waterfall

BundleURL.swift
```diff
@@ -0,0 +1,34 @@
+import CoreML
+
+@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
+//model name : "FindFour.mlmodelc"
+struct BundleURL{
+    static func bundleUrl(modelName: String) -> URL?{
```
kingst: Space between `?{`

BundleURL.swift
```diff
@@ -0,0 +1,34 @@
...
+        var modelcFile: URL
```
kingst: why is this a var? Can it be a `let`

BundleURL.swift
```diff
@@ -0,0 +1,34 @@
...
+        if let documentDirectory =
```
kingst: you can use a `guard` here also since we're always going to need this directory and if we can get it we can't compile

BundleURL.swift
```diff
@@ -0,0 +1,34 @@
...
+                return nil
+            }
+            
+            guard let bundle = Bundle(url: bundleUrl) else {
+                return nil
+            }
+            
+            let modelUrl = bundle.url(forResource: "FindFour", withExtension: "bin")
+            
+            if let compiledUrl =
```
kingst: You should use a `guard` here or return `nil` on the else path because the only way we will return a non null is if everything in this function works

CardScan/Classes/Models.swift
```diff
@@ -99,7 +84,7 @@ class FindFour {
     
     /// Construct a model that automatically loads the model from the app's bundle
     convenience init() {
-        try! self.init(contentsOf: type(of:self).urlOfModelInThisBundle)
+        try! self.init(contentsOf: BundleURL)
```
kingst: Do we need this still? We can just instantiate the class directly using the url we get from your static method

BundleURL.swift
```diff
@@ -3,30 +3,40 @@ import CoreML
 @available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
 //model name : "FindFour.mlmodelc"
 struct BundleURL{
-    static func bundleUrl(modelName: String) -> URL?{
-        var modelcFile: URL
-        if let documentDirectory =
-            try? FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor:nil, create:false) {
-                modelcFile = documentDirectory.appendingPathComponent(modelName)
-            }
+    static func bundleUrl(forResource: String, withExtension: String, modelName: String) -> URL?{
```
kingst: Minor comment, but I'd make `forResource` and `withExtension` the public variable names with something like `resource` and `extension` internally

PR #27, opened 2019-06-27
Change models to optionals

kingst:


jaimejiyepark:
#shipit

jaimejiyepark:
hey btw. is this thread-safe?

kingst:
Definitely _not_ thread safe, but it's all getting called from the machineLearningQueue. Good call, I'll leave a comment before I merge this.

PR #31, opened 2019-07-03
Remove calledDelegate check for cancel and skip

kingst:
As a safety precaution we removed the check to see if the delegate has already been called when the user presses the back button. This means that that method may be invoked multiple times and it could be invoked even after a successful scan, but it does make sure that we always have an out.


jaimejiyepark:
very easy to understand. #shipit!

PR #32, opened 2019-07-05
Added ability to run ScanViewController with custom images

kingst:
This PR will help us with automated testing by adding two new capabilities:
* Feed the machine learning algorithm via a caller-specified `TestingImageDataSource`
* Show the images as they are processing via `showTestingImageView`

With these changes in place, app builders can test this library if they run automated tests on real hardware. The reason this only works with real hardware is because of the way Apple handles camera emulation in their simulators (ie they don't).


CardScan/Classes/ScanViewController.swift
```diff
@@ -31,6 +31,10 @@ import Vision
     @objc func skipButton() -> String
 }
 
+@objc public protocol TestingImageDataSource {
```
jaimejiyepark: Can you give me an example of how this protocol will be used? I read below that it can be used for custom testing but I am still unclear on how it it should be implemented.

kingst: Yeah, checkout `Example/CardScan/ViewController.swift`, which is part of this PR. It shows how to implement this protocol and how to use it.

PR #34, opened 2019-07-10
Add fraud check API call

kingst:
This PR adds a fire-and-forget call to the getbouncer fraudcheck
API. We use this API to monitor the health of our card scanning
library.


jaimejiyepark:
wow looks great!! #shipit

PR #41, opened 2019-07-18
Expose camera start function

kingst:
This PR sets up the base view controller so that it starts in a paused state by default and any classes that inherit from it need to explicitly start the camera by calling the `startCameraPreview` method. The ScanViewController does this during `didLoad` but other VCs can do this later if they choose.


jaimejiyepark:
loooks AMAZING #shipit

PR #42, opened 2019-07-25
Mlmodels as classes

kingst:


CardScan/Classes/GeneratedModels.swift
```diff
@@ -114,9 +133,6 @@ class FindFour {
      - returns: the result of the prediction as FindFourOutput
      */
     func prediction(input: FindFourInput, options: MLPredictionOptions) throws -> FindFourOutput {
-        if AppState.inBackground {
-            throw InBackgroundError.appIsInTheBackground
```
jaimejiyepark: Can you explain to me why you removed this chunk of code?

PR #44, opened 2019-07-25
Add card detection stat

kingst:


jaimejiyepark:
I was able to test scanDetection but how can i test detect/recognize digits?

jaimejiyepark:
besides that, looks good! #shipit

PR #45, opened 2019-07-25
Drop iPhone 6 and older using a device blacklist

kingst:
The GPU on iPhone 6 and iPhone 6 Plus are too slow to support our machine learning models efficiently, so we're dropping them from our supported hardware. We should also investigate iPad and iPod touch later, but for now this is good enough.

I tested this using our Firebase TestLab setup and confirmed that iPhone6 is not supported but all newer devices are.

https://console.firebase.google.com/project/bouncer-card-scan/testlab/histories/bh.1247d8283144895c/matrices/6344922426939065259

### DIFF
--- a/CardScan/Classes/ScanBaseViewController.swift
+++ b/CardScan/Classes/ScanBaseViewController.swift
@@ -92,7 +92,13 @@ import Vision
     
     @objc static public func isCompatible() -> Bool {
         if #available(iOS 11.2, *) {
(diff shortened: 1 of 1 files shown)


kingst:
Here's the image showing iPhone 6 failing (correctly) and everything else working
image

jaimejiyepark:
tested, great, amazing as always. #shipit!

PR #47, opened 2019-07-30
adds expiration date label

justinvallely:
As a user, when a successful scan grabs the expiration date, I'd like to see that value on the resulting view controller along with the card number.


kingst:
#shipit, thanks @justinvallely

PR #51, opened 2019-08-05
Cleanup api

kingst:
Tighten up our `scan_stats` API call, remove unneeded info and make sure that we're collecting the `cards_detected` and pod version fields.


kingst:
Self review #shipit

PR #53, opened 2019-08-08
Added Carthage support

Li-Andy:
- Updated README to include installation instructions for Carthage
- Added RELEASE.md to outline release procedure
- Updated Example/.gitignore to support files needed for Carthage


RELEASE.md
```diff
@@ -0,0 +1,22 @@
+# Release process
+
+## Shipping a new version
+
+Once we have a build on master we want to publish:
+
+1. Run CardScan systems iOS test
```
kingst: should we run `Carthage build` or something else just to make sure that the carthage part still works?

PR #55, opened 2019-08-09
Add scan events protocol

kingst:


kingst:
self commit, but @Li-Andy if you could take a look and make sure that we won't need to do this every time we want to work on a PR that'd be great

Example/.gitignore
```diff
@@ -1,3 +1,4 @@
+CardScan.xcworkspace/xcshareddata
```
Li-Andy: We need to include `xcshareddata` to enable Carthage to find shared schemes

PR #56, opened 2019-08-09
Add a  protocol for scanning extensibility

kingst:
Gives other SDKs the ability to get called back during key scanning events, like recognizing a number and finishing up the scan.


jaimejiyepark:
looking gooooood #shipit

jaimejiyepark:
oh yes! i need toggletorch to be public. thank you! #shipit

PR #57, opened 2019-08-09
Removed nonessential files for Carthage

Li-Andy:
- Remove files in `Target Support Files/` other than `CardScan` target


kingst:
#shipit

PR #58, opened 2019-08-13
Add a helper function to display expiries

kingst:
I'm not 100% sure if we should be using date formatting functions or not with expiry. I believe that it's always MM/YY or MM/YYYY for credit cards, but I could be wrong.


jaimejiyepark:
looking good #shipit

PR #59, opened 2019-08-14
Image fraud check

dxaen:
The first cut of image based fraud check.

### DIFF
--- a/CardScan/Classes/NMS.swift
+++ b/CardScan/Classes/NMS.swift
@@ -1,0 +1,55 @@
+//
+//  NMS.swift
+//  CardScan
+//
+//  Created by Zain on 8/6/19.
+//
+
+import Foundation
+import os.log
+struct NMS{
+    static func hardNMS(subsetBoxes: [[Float]], probs: [Float], iouThreshold: Float, topK: Int, candidateSize: Int) -> [Int] {
+        /** In this project we implement HARD NMS and NOT Soft NMS
+         * I highly recommend checkout SOFT NMS Implementation of Facebook Detectron Framework
+         *
+         *  Args:
+         *  box_scores (N, 5): boxes in corner-form and probabilities.
+         *  iou_threshold: intersection over union threshold.
+         *  top_k: keep top_k results. If k <= 0, keep all the results.
+         *  candidate_size: only consider the candidates with the highest scores.
+         *
+         *  Returns:
+         *  picked: a list of indexes of the kept boxes
+         */
+        
+        
+        let sorted = probs.enumerated().sorted(by: {$0.element > $1.element})
+        var indices = sorted.map{$0.offset}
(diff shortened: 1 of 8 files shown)


CardScan/Classes/NMS.swift
```diff
@@ -0,0 +1,55 @@
...
+         */
+        
+        
+        let sorted = probs.enumerated().sorted(by: {$0.element > $1.element})
+        var indices = sorted.map{$0.offset}
+        var current: Int = 0
+        var currentBox = [Float]()
+        var pickedIndices = [Int]()
+        
+        if(indices.count > 200){
```
kingst: Seems like there's a more "swifty" way to do this

CardScan/Classes/NMS.swift
```diff
@@ -0,0 +1,55 @@
...
+         *  top_k: keep top_k results. If k <= 0, keep all the results.
+         *  candidate_size: only consider the candidates with the highest scores.
+         *
+         *  Returns:
+         *  picked: a list of indexes of the kept boxes
+         */
+        
+        
+        let sorted = probs.enumerated().sorted(by: {$0.element > $1.element})
+        var indices = sorted.map{$0.offset}
```
kingst: maybe something like `let indices = sorted.map { $0.offset }.prefix(200)`

CardScan/Classes/NMS.swift
```diff
@@ -0,0 +1,55 @@
...
+         *  iou_threshold: intersection over union threshold.
+         *  top_k: keep top_k results. If k <= 0, keep all the results.
+         *  candidate_size: only consider the candidates with the highest scores.
+         *
+         *  Returns:
+         *  picked: a list of indexes of the kept boxes
+         */
+        
+        
+        let sorted = probs.enumerated().sorted(by: {$0.element > $1.element})
```
kingst: the sorted part I think can just be `sorted { $0.element > $1.element }` no need for a label or passing it as an argument.

CardScan/Classes/NMSUtil.swift
```diff
@@ -0,0 +1,71 @@
+//
+//  NMSUtil.swift
+//  CardScan
+//
+//  Created by Zain on 8/6/19.
+//
+
+import Foundation
+struct NMSUtils{
+    static func IOUOf(currentBox: [Float], nextBox: [Float]) -> Float{
```
kingst: is it possible to use some sort of apple structure (like CGRect) for this?

kingst: This comment is optional, this code is fine as is for now

kingst: ideally I'd love to see this implemented using an extension to a CGRect that is something like `iou(rect: CGRect)`

PR #60, opened 2019-08-15
Use single delegate object for scanEvents

kingst:
Instead of copying around the delegate object we register `ScanBaseViewController` as a delegate of OCR, which then relays `ScanEvent` events to the same delegate that the caller sets.


dxaen:
Works now. #shipit

PR #61, opened 2019-08-15
added a public getter to a property, made init and a static model initializer public

dxaen:


CardScan/Classes/SsdDetect.swift
```diff
@@ -58,15 +56,21 @@ public struct SsdDetect {
...
+    public init(){
         if SsdDetect.priors == nil{
             SsdDetect.priors = PriorsGen.combinePriors()
         }
+        
+    }
+    
+    public static func initializeModels( contentsOf url: URL) throws {
+        if SsdDetect.ssdModel == nil {
+            SsdDetect.ssdModel = try SSD(contentsOf: url)
```
kingst: convert to try? and add a guard to predict

dxaen: Made the requested changes.

PR #67, opened 2019-08-19
fix priors generation to include more built in types

dxaen:
Removed a lot of garbage code. Added an extension to float for clamping values between 0.0 and 1.0


dxaen:
I can't merge this PR.

PR #68, opened 2019-08-21
return normlized co-ordinates from SSD

dxaen:


kingst:
To keep debugging easy I suggest keeping the rect in image coordinates but convert only the dict to a normalized value

dxaen:
Good point. Let me make the change.

dxaen:
https://github.com/getbouncer/cardscan-ios/pull/68/commits/a58e479b485b3bc5caf40e1781e6b6f915497b20

kingst:
#shipit. Eventually it'd be nice to have toDict(normalized: CGSize) as an extension to CGRect, but for now let's ship this as is.

PR #87, opened 2019-10-09
full screen in storyboard

jaimejiyepark:
You can't change the screen within the VC because it's the presentationStyle. But you can change the presentationStyle in the storyboards.


jaimejiyepark:
> Looks good, much better. How do you plan to test it?

Well if I had admin privileges I could download an older ios version to test with. But for now I just did a manual test with my phone.

kingst:
Can you post a few screen shots? Testing on your phone with and without the change sounds fine.

jaimejiyepark:
**With .fullscreen**
withFull

**Without .fullscreen**
withouFull

PR #91, opened 2019-10-26
Adjust scale before running recognition model

kingst:
This PR uses the bounding box around the card number detection model to adjust the scale for recognition boxes. The underlying assumption is that clusters of 4 digits maintain the same aspect ratio, thus you can get more consistent digit sizes when we do recognition.

This scaling technique only applies to horizontal cards.

### DIFF
--- a/CardScan/Classes/RecognizeNumbers.swift
+++ b/CardScan/Classes/RecognizeNumbers.swift
@@ -24,8 +24,26 @@ struct RecognizeNumbers {
         self.recognizedDigits = Array(repeating: Array(repeating: nil, count: numCols), count: numRows)
     }
     
+    func calculateScale(line: [DetectedBox]) -> Double? {
+        if line.count != 4 {
+            return nil
+        }
+        
+        let numberMinX = line.map({ $0.rect.minX }).min() ?? 0.0
+        let numberMaxX = line.map({ $0.rect.maxX }).max() ?? 0.0
+        let numberWidth = numberMaxX - numberMinX
+        let boxWidth = line.first?.rect.width ?? 1.0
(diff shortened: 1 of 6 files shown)


CardScan/Classes/RecognizeNumbers.swift
```diff
@@ -24,8 +24,26 @@ struct RecognizeNumbers {
...
+    func calculateScale(line: [DetectedBox]) -> Double? {
+        if line.count != 4 {
+            return nil
+        }
+        
+        let numberMinX = line.map({ $0.rect.minX }).min() ?? 0.0
+        let numberMaxX = line.map({ $0.rect.maxX }).max() ?? 0.0
+        let numberWidth = numberMaxX - numberMinX
+        let boxWidth = line.first?.rect.width ?? 1.0
+        let scale = Double(numberWidth * 1.2 / (boxWidth * 4.0))
```
dxaen: Can you please explain this line? Why `1.2`?

dxaen: Also, does it apply to Amex cards?

kingst: I evaluated this empirically. On our validation set from TestOcr the frame prediction recall went from 54% to 64% and on our test set it went from 44% to 65%. The larger increase in testing is likely due to the fact that our testing set has more embossed cards, which appear to benefit more from this optimization.

So in other words, I pulled it out of thin air, but it does seem to work :)

kingst: It only applies to horizontal 16 digit cards. Check out the changes to FindFourOcr.swift, if you expand the hidden parts you'll see that when it calls the `number` method on vertical and amex that `useScale` uses the default value (ie false)

PR #95, opened 2019-11-13
Remove forced unwrap and provide ability to set bundle name

kingst:


CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -133,8 +133,12 @@ import Vision
     }
     
     @objc static public func cameraImage() -> UIImage? {
-        let bundleUrl = Bundle(for: ScanViewController.self).url(forResource: "CardScan", withExtension: "bundle")!
-        let bundle = Bundle(url: bundleUrl)!
+        guard let bundleUrl = Bundle(for: ScanViewController.self).url(forResource: "CardScan", withExtension: "bundle") else {
```
krunk4ever: similar to how you allowed resource and extensionName to be configured in `BundleURL`, would it be possible to have bundle resource loading go through that same path with a customiziable `bundleName`, `extensionName`?

kingst: Yeah, agreed. I think we came up with something better and also provide the ability to set a bundle directly if you choose to do so. Check out the latest commit.

CardScan/Classes/BundleURL.swift
```diff
@@ -1,10 +1,17 @@
...
-
-@available(iOS 11.2, *)
-struct BundleURL{
-    static func compiledModel(forResource: String, withExtension: String) -> URL? {
+public struct BundleURL {
+    // If you change the bundle name make sure to set these before
+    // initializing the library
+    public static var bundleName = "CardScan"
+    public static var extensionName = "bundle"
+    public static var cardScanBundle: Bundle?
```
jaimejiyepark: Are you setting these values statically to have default values? So it'll always look for CardScan.bundle if not changed. Why not it be empty so it can be filled it like ```BundleUrl(bundleName:"Whatever the user puts in", extenstionName: "whatever")```

kingst: The goal is to let the app that embeds our SDK specify the bundle name or the actual bundle itself to give them more flexibility in their build system. Thus, BundleURL shouldn't even get called externally (except for testing), it's only internal so we're a little constrained, although I do agree with your sentiment.

jaimejiyepark:
lgtm #shipit

PR #97, opened 2019-11-20
exposing torch functions

jaimejiyepark:
* Exposes torch state and torch availability through function calls
* Modify torch level intensity through initial ScanViewController set up or through function `setTorchLevel`
* If no torch level is set, then it is defaulted to 1.0

### DIFF
--- a/CardScan/Classes/Torch.swift
+++ b/CardScan/Classes/Torch.swift
@@ -9,6 +9,11 @@ struct Torch {
     let device: AVCaptureDevice?
     var state: State
     var lastStateChange: Date
+    var level: Float {
+        didSet {
+            self.setTorchMode()
+        }
+    }
     
     init(device: AVCaptureDevice) {
         self.state = .off
@@ -19,22 +24,12 @@ struct Torch {
         } else {
             self.device = nil
         }
+        self.level = 1.0
     }
     
     mutating func toggle() {
         self.state = self.state == .on ? .off : .on
-
-        do {
-            try self.device?.lockForConfiguration()
-            if self.state == .on {
-                do { try self.device?.setTorchModeOn(level: 0.5) } catch { print("could not set torch mode on") }
-            } else {
-                self.device?.torchMode = .off
-            }
-            self.device?.unlockForConfiguration()
-        } catch {
-            print("error setting torch level")
-        }
+        self.setTorchMode()
     }
     
     mutating func luma(_ value: Double) {
@@ -52,18 +47,22 @@ struct Torch {
         if self.state != newState {
             self.lastStateChange = Date()
             self.state = newState
-            
-            do {
-                try self.device?.lockForConfiguration()
-                if self.state == .on {
-                    do { try self.device?.setTorchModeOn(level: 0.005) } catch { print("could not set torch mode on") }
-                } else {
-                    self.device?.torchMode = .off
-                }
-                self.device?.unlockForConfiguration()
-            } catch {
-                print("error setting torch level")
+            self.setTorchMode(level: 0.005)
+        }
+    }
+    
+    func setTorchMode(level: Float? = nil) {
+        let setLevel = level ?? self.level
+        do {
+            try self.device?.lockForConfiguration()
+            if self.state == .on {
+                do { try self.device?.setTorchModeOn(level: setLevel) } catch { print("could not set torch mode on") }
+            } else {
+                self.device?.torchMode = .off
             }
+            self.device?.unlockForConfiguration()
+        } catch {
+            print("error setting torch level")
         }
     }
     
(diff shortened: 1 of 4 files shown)


CardScan/Classes/Torch.swift
```diff
@@ -19,22 +24,12 @@ struct Torch {
...
-                self.device?.torchMode = .off
-            }
-            self.device?.unlockForConfiguration()
-        } catch {
-            print("error setting torch level")
-        }
+        self.setTorchMode()
     }
     
     mutating func luma(_ value: Double) {
```
kingst: does this function get used anywhere? If so, where?

CardScan/Classes/Torch.swift
```diff
@@ -9,6 +9,11 @@ struct Torch {
     let device: AVCaptureDevice?
     var state: State
     var lastStateChange: Date
+    var level: Float {
+        didSet {
```
kingst: I don't think that we've used property observers anywhere else, I don't think we should introduce them for this change

CardScan/Classes/VideoFeed.swift
```diff
@@ -176,10 +176,24 @@ class VideoFeed {
         videoDevice.unlockForConfiguration()
     }
     
+    //MARK: -Torch Logic
     func toggleTorch() {
         self.torch?.toggle()
     }
     
+    func isTorchOn() -> Bool {
+        return self.torch?.state ==  Torch.State.on ? true : false
```
kingst: Isn't the `? :` redundant here?

CardScan/Classes/VideoFeed.swift
```diff
@@ -176,10 +176,24 @@ class VideoFeed {
...
+    }
+    
+    func hasTorch() -> Bool {
+        return self.torch?.device?.hasTorch ?? false
```
kingst: I don't know if `hasTorch` or `hasTorchAndIsAvailable` is better, do you have an opinion?

CardScan/Classes/VideoFeed.swift
jaimejiyepark: well isnt the state an enum and not a bool? so if its not on, return False

CardScan/Classes/VideoFeed.swift
jaimejiyepark: hmmmm... i guess hasTorch&isAvailable is the safer route. So I'll go with that.

CardScan/Classes/Torch.swift
jaimejiyepark: i looked and it doesnt. you want me to get rid of it?

CardScan/Classes/Torch.swift
jaimejiyepark: well the reason I used the observer is because once the property is set, it wont be set into torchMode until the `setTorchMode` function is called. so without this, the intial torch level will be 1.0, then when set to some value, it will only be changed after a toggle.

CardScan/Classes/Torch.swift
jaimejiyepark: you're talking about `luma` right?

kingst: yeah luma. Let's kill it.

CardScan/Classes/VideoFeed.swift
kingst: yeah, you need the `self.torch?.state == Torch.state.on` part, just not the  `? true : false` part

CardScan/Classes/VideoFeed.swift
kingst: 👍

CardScan/Classes/Torch.swift
kingst: wouldn't you always have the level when you turn the torch on for the first time? I'm probably just not following something.

CardScan/Classes/VideoFeed.swift
jaimejiyepark: OOOOO yeap yeap yeap youre right

CardScan/Classes/Torch.swift
jaimejiyepark: hmmm... you're right. let me update.

CardScan/Classes/VideoFeed.swift
```diff
@@ -176,10 +176,28 @@ class VideoFeed {
...
     func toggleTorch() {
         self.torch?.toggle()
     }
     
+    func isTorchOn() -> Bool {
+        return self.torch?.state == Torch.State.on 
+    }
+    
+    func hasTorchAndIsAvailable() -> Bool {
+        if let _ = self.torch?.device?.hasTorch, let _ = self.torch?.device?.isTorchAvailable {
```
kingst: Nit: I think a more direct way to express this is:
```
let hasTorch = self.torch?.device?.hasTorch ?? false
let isTorchAvailable = self.torch?.device?.isTorchAvailable ?? false
return hasTorch && isTorchAvailable
```

PR #99, opened 2019-11-26
Adapt to iOS 9

igor-makarov:
This PR adapts the podspec to iOS 9. Functionality is not affected as it's still gated to iOS 11.2.

### DIFF
--- a/CardScan.podspec
+++ b/CardScan.podspec
@@ -13,7 +13,7 @@ CardScan is a library for scanning credit cards.
   # lint warning, who knows
   #s.social_media_url = 'https://twitter.com/stk'
   s.default_subspec = 'Core'
-  s.ios.deployment_target = '10.0'
+  s.ios.deployment_target = '9.0'
   s.swift_version = '4.2'
 
   s.subspec 'Core' do |core|
--- a/CardScan/Classes/NMS.swift
+++ b/CardScan/Classes/NMS.swift
@@ -32,7 +32,9 @@ struct NMS{
         if(indices.count > 200){
             // TODO Fix This
             indices = Array(indices[0..<200])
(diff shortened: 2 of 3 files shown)


kingst:
Your changes look pretty reasonable, but is there a way you could gate the feature on your end? I don't know if you can put the same `if` statement around the import / use of CardScan.

I'm open to this change, but I want to be careful that we're not signing up for a big maintenance burden since our goal is to support fewer OSes instead of more.

igor-makarov:
There's no way to gate this elsewhere. CocoaPods doesn't let you install pods with a higher deployment target.

Note that the proposed change doesn't actually enable any functionality. Nothing will be available on 9. This is just to get the library to build.

kingst:
Yes, I know that it doesn't affect the functionality, the reason that I'm hesitant is that we don't have a great way to test out iOS 9 and if we lower our development target we need to support it.

Have you tried it out on an iOS 9 device and is it working as expected?

igor-makarov:
Not yet, we only have one iOS 9 device at the company 😂

It will get tested, however I honestly don't see what is there to test beyond availability gating.

Have you considered using a free CI service for tests? They usually have older versions available.

kingst:
We use a paid service, Firebase TestLab, but they don't have any OS version that go back that far!

I'll tell you what, this seems like a reasonable change, we'll do it for you but please do let us know if you bump your minimum version and we'll do the same since you're the only one who has asked for this so far :)

igor-makarov:
Cool! 

Let me know when the pod is up so I can update the local source.

Why are you using a paid service? This is an open source library, you can get free macOS machines from Travis, Circle, Azure...  Probably more. Why not?

kingst:
We'll put this in our next release, which should come out this week

PR #102, opened 2019-12-03
mod testing image protocol/ add squared image

jaimejiyepark:
[RFC] 
* set a imageFullScreen bool that can only be set by the `TestingImageDataSource` protocol implementer
* `captureOutputWork` checks:
    * if testing image exists
        * if full screen image -> then get squared and full image.
        * if cropped -> send straight to ocr
   * if frames from camera
     * get squared and full image


jaimejiyepark:
Did not implement cropping the square in OCR to a credit card size yet since I want to get this portion done first.

jaimejiyepark:
* removed the objc attribute to use optionals
* made an extension to `scanBaseviewController` that returns an optional CGImage array from the `testingImageDataSource`
* added logic to `captureOutputWork` if testing images exist

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -18,7 +19,7 @@ import Vision
     }
     
     
-    @objc public weak var testingImageDataSource: TestingImageDataSource?
+    public var testingImageDataSource: TestingImageDataSource?
```
kingst: I think we need to leave the `weak` in here to avoid a memory leak, but I'm not 100% sure

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -440,15 +418,18 @@ import Vision
             return
         }
         
+        let testImgArr: [CGImage?] = self.testingImagesSizeCheck()
```
kingst: from a stylistic perspective, I'd much rather see you return a tuple and assign named variables. eg:

`let (squareCardImage, fullCardImage) = self.testingImageSizeCheck()`

kingst: can we name name this function something else? Maybe `testingImagesWithSizeCheck`

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -440,15 +418,18 @@ import Vision
...
+        
         // we allow apps that integrate to supply their own sequence of images
         // for use in testing
-        let image = self.testingImageDataSource?.nextImage() ?? squareCardImage
+        let squareCardImg: CGImage  = testImgArr[0] ?? squareCardImage
```
kingst: it's more swift-y to name this variable squareCardImage (the swift people usually don't want you to use abbreviations)

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -440,15 +418,18 @@ import Vision
...
+        let fullCardImg: CGImage = testImgArr[1] ?? fullCardImage
```
kingst: Same here

CardScan/Classes/TestingImageExtension.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  TestingImageExtension.swift
+//  CardScan
+//
+//  Created by Jaime Park on 12/3/19.
+//
+
+import Foundation
+
+extension ScanBaseViewController {
```
kingst: Can we extend the protocol instead? This method, logically, seems to go with the protocol. It makes sense to put it in an extension, if that works with swift (which I think it does)

CardScan/Classes/TestingImageExtension.swift
```diff
@@ -0,0 +1,24 @@
...
+//  TestingImageExtension.swift
+//  CardScan
+//
+//  Created by Jaime Park on 12/3/19.
+//
+
+import Foundation
+
+extension ScanBaseViewController {    
+    func testingImagesSizeCheck() -> [CGImage?]{
```
kingst: Is there a cleaner way to handle this. I think what you have looks correct, but it's a little hard to follow / more verbose that it might have to be. Play around with it and see if you can come up with anything that you think looks cleaner

Example/CardScan/ViewController.swift
```diff
@@ -10,7 +10,8 @@ import UIKit
 import CardScan
 
 class ViewController: UIViewController, ScanEvents, ScanDelegate, FullScanStringsDataSource, TestingImageDataSource {
-
+    var imageIsFullScreen: Bool? = nil
```
kingst: can we _not_ define this? Our goal is to make sure that the people who are already using this don't need to change their code, so ideally we wouldn't need to make any changes to this function. It might not be possible, but we should try.

CardScan/Classes/TestingImageExtension.swift
jaimejiyepark: oooh this is what you meant by your message. got it!

CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: it wont let me put the `weak` in there :-/ . i get the errors ```'weak' must not be applied to non-class-bound 'TestingImageDataSource'```.

jaimejiyepark:
I'll get on the end-end test once we approve the protocol design

CardScan/Classes/ScanBaseViewController.swift
kingst: we should understand this more clearly. the "safe" thing to do here is to add the @objc and weak back and be done with it, but if you want to make a change it's going to require some deeper analysis

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -520,8 +497,7 @@ import Vision
         
         let rect = CGRect(x: cx - width / 2.0, y: cy - height / 2.0, width: width, height: height)
         
-        self.currentImageRect = rect
```
kingst: what is this change for?

Example/CardScan/ViewController.swift
```diff
@@ -24,10 +23,29 @@ class ViewController: UIViewController, ScanEvents, ScanDelegate, FullScanString
...
     
-    func nextImage() -> CGImage? {
-        let nextImage = self.currentTestImages?.first
+    func nextSquareAndFullImage() -> (CGImage, CGImage)? {
+        guard let fullCardImage = self.currentTestImages?.first else {
+            print("could not get full size test image")
+            return nil
+        }
+        
+        let cropToSquareImage: CGImage? = {
```
kingst: why are you doing this with a closure?

CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: this is from one of my first commits, i just got rid of it since it doesnt looks like we're using it anywhere

Example/CardScan/ViewController.swift
jaimejiyepark: i just did it so i could encapsulate it. no strong feelings about it. should i change it?

CardScan/Classes/ScanBaseViewController.swift
kingst: Let's just leave it in for now, we can do another PR that removes this that's separate.

CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: im going to see if theres a workaround in the next commit to just add the @objc back. it's all because of that optional return

jaimejiyepark:
`AnyObject` allows the protocol to only be conformed by class-types which allows us to put in that `weak` for the delegate.

https://developer.apple.com/documentation/swift/anyobject

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -438,13 +413,13 @@ import Vision
...
         // for use in testing
-        //let image = self.testingImageDataSource?.nextImage() ?? squareCardImage
+        let squareAndFullCardImage = self.testingImageDataSource?.nextSquareAndFullImage() ?? (squareCardImage, fullCardImage)
         
         if #available(iOS 11.2, *) {
             if self.scanQrCode {
                 self.blockingQrModel(pixelBuffer: pixelBuffer)
             } else {
-                self.blockingOcrModel(squareCardImage: squareCardImage, fullCardImage: fullCardImage)
+                self.blockingOcrModel(squareCardImage: squareAndFullCardImage.0, fullCardImage: squareAndFullCardImage.1)
```
kingst: naming individual tuple elements here doesn't seem great people would need to know that .0 is square and .1 is full. For squareAndFullCardImage why not have variables directly, something like `let (square, full) = self.testingImageDataSource...`

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -521,7 +496,6 @@ import Vision
         let rect = CGRect(x: cx - width / 2.0, y: cy - height / 2.0, width: width, height: height)
         
         self.currentImageRect = rect
-
```
kingst: Can you remove whitespace changes from here and the few other places that they show up?

jaimejiyepark:
I would argue that its actually not hacky. 
https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#//apple_ref/doc/uid/TP40014097-CH25-ID267
The first portion of the delegate section shows an example of a protocol inheriting `AnyObject`

PR #103, opened 2019-12-03
[RFC] First cut at sending square and full images to the ML pipeline

kingst:
I have only compiled this, I haven't even run it yet. It's just to get feedback on the direction I'm taking it.


kingst:
DANGER -- I'm going to merge this to make it easy for Jaime and I to both work on this same chunk of code -- expect a follow on PR that fixes all of the bugs that I'm about to add

PR #104, opened 2019-12-04
Update the ScanCard protocol, more robust cropping

kingst:
These include tweaks for sending the info we need for verify + adjusting the cropping algorithm so that it always crops.


jaimejiyepark:
lgtm #ship

PR #107, opened 2019-12-16
[RFC] bin rules

jaimejiyepark:


jaimejiyepark:
Adam and I decided to follow the Stripe-android code for the bin rules since the the bin set is the most similar to the IIN's listed on wiki 
https://en.wikipedia.org/wiki/Payment_card_number#Issuer_identification_number_(IIN)

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -1,6 +1,25 @@
 import Foundation
 
 public struct CreditCardUtils {
+    static let cvcLengthAmericanExpress = 4
```
kingst: are the cvc lengths getting used anywhere?

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -1,6 +1,25 @@
...
+    static let cvcLength = 3
+    
+    static let maxPanLength = 16
```
kingst: according to the iOS code some of the pans for visa are 14, do they not track that in the Android side?

jaimejiyepark: https://github.com/stripe/stripe-android/blob/7f78931e92c81257fe7443a3dd558273e9b45c44/stripe/src/main/java/com/stripe/android/CardUtils.kt#L107

they only check for diners club and amex, otherwise standard length

CardScan/Classes/CreditCardUtils.swift
jaimejiyepark: no but i thought maybe we'd use them later?

kingst: Maybe, but let's remove them for now and add them back when we need them.

CardScan/Classes/CreditCardUtils.swift
kingst: ok, as long as we're consistent I'm fine with it

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -25,15 +41,12 @@ public struct CreditCardUtils {
     }
     
     public static func isValidBin(number: String) -> Bool {
```
awushensky: In the android code, I used the `determineCardNetwork` method to check for validity. This is more readable an extensible than checking for each possible card:
https://github.com/getbouncer/cardscan-android/blob/master/cardscan-base/src/main/java/com/getbouncer/cardscan/base/CreditCardUtils.kt#L125

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -25,15 +41,12 @@ public struct CreditCardUtils {
...
-        return isAmex(number: number) || isDiscover(number: number) || isVisa(number: number) || isMastercard(number: number) || isUnionPay(number: number)
+        return isAmex(number: number) || isDiscover(number: number) || isVisa(number: number) || isMastercard(number: number) || isUnionPay(number: number) || isJcb(number: number) || isDinersClub(number: number)
     }
     
     public static func isAmex(number: String) -> Bool {
```
awushensky: Can we get rid of these methods? Instead of specific `isNetwork` methods, we should simplify the methods using these.

For example, if we are using these methods to set an icon
```
if (CreditCardUtils.isAmex()) {
   setIcon(amexIcon)
} else if (CreditCardUtils.isUnionPay()) {
  setIcon(unionPay)
} else if ...
```

we should instead create a method in CreditCardUtils that returns the desired icon.
```
switch (determineCardNetwork(number)) {
    case VISA -> visaIcon
    case AMEX -> amexIcon
    ...
}
```

awushensky: Centralizing into one `determineCardNetwork` method also keeps all the prefix logic in one place instead of spread out around multiple methods

see https://github.com/getbouncer/cardscan-android/blob/master/cardscan-base/src/main/java/com/getbouncer/cardscan/base/CreditCardUtils.kt#L159

awushensky: this _also_ lets us more easily deal with prioritizing discover over unionpay when BINs overlap.

Example/CardScan_ExampleTests/CardScan_CardUtilsTests.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  CardScan_CardUtilsTests.swift
+//  CardScan_ExampleTests
+//
+//  Created by Jaime Park on 12/16/19.
+//  Copyright © 2019 CocoaPods. All rights reserved.
+//
+
+import XCTest
+@testable import CardScan
```
awushensky: NICE! thanks for adding tests for this!

jaimejiyepark:
thank you @awushensky for all the feedback! let me fixes these changes

PR #108, opened 2019-12-19
card number/ bin validity check

jaimejiyepark:
note:
* changed all argument names to `cardNumber` to stay consistent

### DIFF
--- a/CardScan/Classes/CreditCardUtils.swift
+++ b/CardScan/Classes/CreditCardUtils.swift
@@ -1,29 +1,43 @@
 import Foundation
 
+public enum CardNetwork: String {
+    case VISA = "Visa"
+    case MASTERCARD = "Mastercard"
+    case AMEX = "Amex"
+    case DISCOVER = "Discover"
+    case UNIONPAY = "Union Pay"
+    case JCB = "Jcb"
+    case DINERSCLUB = "Diners Club"
+    case UNKNOWN = "Unknown"
+}
+
 public struct CreditCardUtils {
     static let maxPanLength = 16
     static let maxPanLengthAmericanExpress = 15
     static let maxPanLengthDinersClub = 14
 
-    static let prefixesAmericanExpress = ["34", "37"]
-    static let prefixesDinersClub = ["300", "301", "302", "303", "304", "305", "309", "36", "38", "39"]
-    static let prefixesDiscover = ["6011", "64", "65"]
-    static let prefixesJcb = ["35"]
-    static let prefixesMastercard = ["2221", "2222", "2223", "2224", "2225", "2226",
+    private static let prefixesAmericanExpress = ["34", "37"]
+    private static let prefixesDinersClub = ["300", "301", "302", "303", "304", "305", "309", "36", "38", "39"]
+    private static let prefixesDiscover = ["6011", "64", "65"]
+    private static let prefixesJcb = ["35"]
+    private static let prefixesMastercard = ["2221", "2222", "2223", "2224", "2225", "2226",
                                      "2227", "2228", "2229", "223", "224", "225", "226",
                                      "227", "228", "229", "23", "24", "25", "26", "270",
                                      "271", "2720", "50", "51", "52", "53", "54", "55",
                                      "67"]
-    static let prefixesUnionPay = ["62"]
-    static let prefixesVisa = ["4"]
+    private static let prefixesUnionPay = ["62"]
+    private static let prefixesVisa = ["4"]
+    
+    
+    public static func isValidNumber(cardNumber: String) -> Bool {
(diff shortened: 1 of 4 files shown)


jaimejiyepark:
I changed `luhnCheck` and `isValidBin` to non-public since they are functions used in conjunction with each other in the greater `isValidNumber` function. I noticed that `luhnCheck` was being used in `RecognizeNumbers.swift` with checking the `number.count` which is what `isValidNumber` does. Also, `isValidBin` in the context of cards while in `CardVerify` we use it to get the card icon images. To remedy that I made a function `determineCardNetwork` public for use. I also followed Adam's android code pretty closely.

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -1,29 +1,43 @@
 import Foundation
 
+public enum CardNetwork: String {
```
kingst: does this need to be public?

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -1,29 +1,43 @@
...
+    }
     
     // https://en.wikipedia.org/wiki/Luhn_algorithm
     // assume 16 digits are for MC and Visa (start with 4, 5) and 15 is for Amex
     // which starts with 3
-    public static func luhnCheck(_ cardNumber: String) -> Bool {
-        if cardNumber.count == 0 {
-            return false
-        } else if !isValidBin(number: cardNumber) {
+    static func isValidLuhnNumber(cardNumber: String) -> Bool {
```
kingst: You're taking a function that used to be public and making it not public. If people who use our library use this function they're going to be upset that they need to change code now.

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -40,68 +54,76 @@ public struct CreditCardUtils {
         return sum % 10 == 0
     }
     
-    public static func isValidBin(number: String) -> Bool {
```
kingst: Same here

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -40,68 +54,76 @@ public struct CreditCardUtils {
...
     }
     
-    public static func isAmex(number: String) -> Bool {
-        let prefix = String(number.prefix(2))
-        return number.count == maxPanLengthAmericanExpress && prefixesAmericanExpress.contains(prefix)
+    static func isValidLength(cardNumber: String) -> Bool {
+        return isValidLength(cardNumber: cardNumber, network: determineCardNetwork(cardNumber: cardNumber))
     }
-    
-    public static func isUnionPay(number: String) -> Bool {
```
kingst: Same with all of the isSomeBank functions

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -40,68 +54,76 @@ public struct CreditCardUtils {
...
+            return cardNumberLength == maxPanLengthDinersClub
+        default:
+            return cardNumberLength == maxPanLength
+        }
     }
     
-    public static func isJcb(number: String) -> Bool {
-        let prefix = String(number.prefix(2))
-        return number.count == maxPanLength && prefixesJcb.contains(prefix)
+    public static func determineCardNetwork(cardNumber: String) -> CardNetwork {
```
kingst: Ok, I see, ignore my comment above

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -40,68 +54,76 @@ public struct CreditCardUtils {
...
+    static func hasAnyPrefix( cardNumber: String, prefixes: [String] ) -> Bool {
+        return prefixes.filter { cardNumber.hasPrefix($0) }.count > 0
     }
     
-    public static func format(number: String) -> String {
-        if number.count == 16 {
-            return format16(number: number)
-        } else if number.count == 15 {
-            return format15(number: number)
+    public static func format(cardNumber: String) -> String {
```
kingst: if people are using this function they have to change their source code after this change, which we should avoid for changes like this

jaimejiyepark:
No it doesn't change anything written in the documentation. It only changes the internal interactions.

jaimejiyepark:
okay real quick though, can i add the `isValidCVV` and `isValidExpirationDate` logic to this too. It's originally in `CardVerify`. Doesn't affect the `CardVerify` docs

CardScan/Classes/CreditCardUtils.swift
jaimejiyepark: its public so it can be used in `CardVerify`

jaimejiyepark:
Ill just make that a separate PR

PR #109, opened 2019-12-23
added cvv and exp date func, added md comments

jaimejiyepark:
* added additional cvv and expiration checks
* added markdown comments since I saw documentation on yours and I thought it was a good idea to start documenting. Also the markdown comments will appear formatted when user's implement functions which is nice


jaimejiyepark:
Should I get Sam to look over it as well?

kingst:
Looks great, thanks for doing this. #shipit

PR #111, opened 2020-01-16
providing client-side function with current frames number and predicted number

jaimejiyepark:
* Exposed the current frame's card number to the return of `OCR`'s `performWithErrorCorrection()`
* The function `uponPredictedCardNumber()` allows clients to handle the case where the predicted number and expected number do not align with their own implementation

### DIFF
--- a/CardScan/Classes/Ocr.swift
+++ b/CardScan/Classes/Ocr.swift
@@ -60,7 +60,7 @@ public class Ocr {
     }
     
     @available(iOS 11.2, *)
-    public func performWithErrorCorrection(for croppedCardImage: CGImage, squareCardImage: CGImage, fullCardImage: CGImage, predictedNumberPredicate: (String?, String) -> Bool ) -> (String?, Expiry?, Bool, Bool) {
+    public func performWithErrorCorrection(for croppedCardImage: CGImage, squareCardImage: CGImage, fullCardImage: CGImage, predictedNumberPredicate: (String?, String) -> Bool = { _, _ in true } ) -> (String?, Expiry?, Bool, Bool) {
         let number = self.perform(croppedCardImage: croppedCardImage, squareCardImage: squareCardImage, fullCardImage: fullCardImage, predictedNumberPredicate: predictedNumberPredicate)
 
         if self.firstResult == nil && number != nil {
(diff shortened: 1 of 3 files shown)


CardScan/Classes/Ocr.swift
```diff
@@ -60,8 +60,8 @@ public class Ocr {
     }
     
     @available(iOS 11.2, *)
-    public func performWithErrorCorrection(for croppedCardImage: CGImage, squareCardImage: CGImage, fullCardImage: CGImage) -> (String?, Expiry?, Bool, Bool) {
-        let number = self.perform(croppedCardImage: croppedCardImage, squareCardImage: squareCardImage, fullCardImage: fullCardImage)
+    public func performWithErrorCorrection(for croppedCardImage: CGImage, squareCardImage: CGImage, fullCardImage: CGImage, predictedNumberPredicate: (String?, String) -> Bool ) -> (String?, Expiry?, Bool, Bool) {
```
kingst: Can you look at if it's possible to have a default parameter for the `predictedNumbersPredicate`, something like `{ _, _ in true}`

I don't know if this is possible with Swift or if there are any subtleties, so we should be thorough if we do go down this route

CardScan/Classes/Ocr.swift
```diff
@@ -90,9 +90,17 @@ public class Ocr {
     }
     
     @available(iOS 11.2, *)
-    public func perform(croppedCardImage: CGImage, squareCardImage: CGImage?, fullCardImage: CGImage?) -> String? {
+    public func perform(croppedCardImage: CGImage, squareCardImage: CGImage?, fullCardImage: CGImage?, predictedNumberPredicate: (String? , String) -> Bool ) -> String? {
```
kingst: same here for default value

CardScan/Classes/ScanViewController.swift
```diff
@@ -304,6 +304,12 @@ import UIKit
         }
     }
     
+    public override func usePredictedCardNumber(predictedNumber: String?, currentFrameNumber: String) -> Bool {
+        print("predicted number: \(predictedNumber)")
```
kingst: can you remove these print statements?

CardScan/Classes/ScanViewController.swift
```diff
@@ -304,6 +304,12 @@ import UIKit
         }
     }
     
+    public override func usePredictedCardNumber(predictedNumber: String?, currentFrameNumber: String) -> Bool {
```
kingst: maybe change the name to `useCurrentFrameNumber` since I think that the `predictedNumber` here is the error corrected number

CardScan/Classes/ScanViewController.swift
jaimejiyepark: yup

CardScan/Classes/ScanViewController.swift
jaimejiyepark: you got it

PR #117, opened 2020-02-03
added framedetected protocol to grab non-numbered frames

jaimejiyepark:


CardScan/Classes/Ocr.swift
```diff
@@ -110,23 +110,24 @@ public class Ocr {
...
+            
+                self.scanStats.algorithm = findFour.algorithm
+                self.updateStats(model: findFour.modelString, boxes: findFour.predictedBoxes, image: croppedCardImage, number: number, cvvBoxes: findFour.cvvBoxes)
+                return number
             }
             
-            self.scanStats.algorithm = findFour.algorithm
-            self.updateStats(model: findFour.modelString, boxes: findFour.predictedBoxes, image: croppedCardImage, number: number, cvvBoxes: findFour.cvvBoxes)
-            return number
+            self.scanEventsDelegate?.onFrameDetected(number: number, expiry: findFour.expiry, numberBoundingBox: numberBoundingBox, expiryBoundingBox: findFour.expiryBoxes.first, croppedCardSize: croppedCardSize, squareCardImage: squareCardImage, fullCardImage: fullCardImage)
```
kingst: I would have assumed that this would go in the else side of things. So in other words we either detect a number or a frame but never both.

PR #120, opened 2020-02-04
adding CardScan pod scheme for carthage build

jaimejiyepark:
CardScan scheme in Pods was not clicked as `shared` so this PR is to share it for carthage build


jaimejiyepark:
I tested it manually: Tried to build carthage, didn't build, checked the schemes, noticed that CardScan in Pods wasn't clicked, selected it, rebuilt, and kapow carthage built.

PR #121, opened 2020-02-06
Obj detect v003

dxaen:
Adds a new class to the list of classes we detect. But we adjust the labels of the classes we send to the server to keep the server implementation consistent.

### DIFF
--- a/CardScan/Classes/DetectedSSDBox.swift
+++ b/CardScan/Classes/DetectedSSDBox.swift
@@ -20,22 +20,23 @@ struct DetectedSSDBox {
         let YMin_ = YMin * Double(imageSize.height)
         let YMax_ = YMax * Double(imageSize.height)
         
-        // The model ouputs labels that are off by 1
-        // compared to the previous versions and this line
-        // serves to retain the consistency. This label
-        // correction should be removed in the future.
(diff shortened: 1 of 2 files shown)


CardScan/Classes/DetectedSSDBox.swift
```diff
@@ -19,19 +19,24 @@ struct DetectedSSDBox {
         let XMax_ = XMax * Double(imageSize.width)
         let YMin_ = YMin * Double(imageSize.height)
         let YMax_ = YMax * Double(imageSize.height)
-
+
```
jaimejiyepark: I feel like @kingst is going to nit these spaces

jaimejiyepark:
lgtm

CardScan/Classes/DetectedSSDBox.swift
dxaen: Ok. Let me fix this real quick.

kingst: nit: let's remove this whitespace change

dxaen: Sure. Also, I tested it locally and the labels were correct.

PR #125, opened 2020-02-08
Update expiry thresholds for better accuracy

kingst:


CardScan/Classes/Expiry.swift
```diff
@@ -24,7 +24,7 @@ public struct Expiry: Hashable {
     
     @available(iOS 11.2, *)
     static func from(image: CGImage, within rect: CGRect) -> Expiry? {
-        guard let digits = RecognizedDigits.from(image: image, within: rect) else {
+        guard let digits = RecognizedDigits.from(image: image, within: rect, minConfidence: 0.65) else {
```
jaimejiyepark: is this an arbitrary value?

kingst: no

PR #126, opened 2020-02-11
[RFC] protocol to extract ocr_result

jaimejiyepark:
Overloaded `onScanComplete` protocol to extract ocr_result for CardVerify to add to `scan` object


kingst:
Did you consider putting this functionality here: https://github.com/getbouncer/cardverify-ios/blob/master/CardVerify/Classes/ScanBaseViewController%2BVerify.swift

All of the data is already here and it feels like a more natural way to handle it since it piggy backs off of our existing mechanism for distilling OCR predictions into something that we surface.

jaimejiyepark:
ahhh! thats so much cleaner than what i wrote. yea that makes way more sense

PR #129, opened 2020-02-14
Provide the ability to override the ML model in ScanBaseViewController

kingst:
Classes that subclass ScanBaseViewController can still use the image processing pipeline but swap out the underlying ML pipeline by overriding the `blockingMlModel` method


jaimejiyepark:
lgtm

PR #133, opened 2020-03-04
Fixing landscape/portrait bug by forcing portrait 

jaimejiyepark:
#### Overview:
To mitigate landscape / portrait mode snafu, I initially started by mocking the UI during transition. Did not work. Then I tried to do the navigation controller overriding `supportedInterfaceOrientation` method that is all over google, the main solution. Also did not work. So the solution I came up with is `1. Lock scanViewController to portrait before transition` and `2. Do not allow orientation transitions once in scanViewController` using `appDelegate` and setting the `UIDevice.current.orientation` to portrait. 

#### To Test:
1. Make sure your device lock is not on
2. Click `Present Modally` button
3. Go in either portrait or landscape mode before clicking `To ScanViewController`
4. Rotate phone like a madman
    * Notice that view controller locks into portrait and will not rotate

#### How it works:
1. ScanViewController life cycle state stored in CardScan. 
    * Can't store in ScanViewController to access in AppDelegate so stored in CardScan
2. ScanViewController's `viewWillAppear` locks _DEVICE_ orientation to always load as portrait 
3. If any rotations are made, appDelegate ` func application(supportedInterfaceOrientationsFor window)` is called
    * Takes scanViewController life cycle state and locks _APP_ orientation to portrait.
#### Notes and Questions:
* Not ideal to make a state flag for scanViewController but it's the only way I can think of without doing nested if statements of "is it navigation controller? what is current vc? is it a scan view controller? is it tab bar controller?" . 
* `supportedInterfaceOrientation` is the range of orientations it supports, `preferredInterfaceOrientation` is a single orientation it should take within supported. 
* `supportedInterfaceOrientations` runs only if `shouldAutoRotate` is true. [Stated in first paragraph of discussion section](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621435-supportedinterfaceorientations).


Example/CardScan/Base.lproj/Main.storyboard
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14868" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="vXZ-lx-hvc">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="15702" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="vXZ-lx-hvc">
```
jaimejiyepark: Added new button to `present modally` and two new view controllers in example

jaimejiyepark:
Changed the isAppearing flag to scanBase

jaimejiyepark:
I rebased earlier but it gave me merge-conflicts since this PR was put up before the bundle changes were pushed to master. Got rid of ```NavViewController```  and I forgot to take the ```ScanViewControllerState``` out of bundle. I think all the changes made it hard to see that I I added the isAppearing flag to ```ScanBaseViewController```.

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -29,6 +29,7 @@ public protocol TestingImageDataSource: AnyObject {
     
     public var scanEventsDelegate: ScanEvents?
     
+    static public var isAppearing = false
```
jaimejiyepark: here!

kingst: got it!

Example/CardScan/ViewController.swift
```diff
@@ -212,4 +212,3 @@ class ViewController: UIViewController, ScanEvents, ScanDelegate, FullScanString
         self.present(vc, animated: true)
     }
 }
-
```
kingst: alright, now you're just trolling me :)

PR #135, opened 2020-03-04
change cardscan to csbundle

jaimejiyepark:


jaimejiyepark:
carthage and firebase tests came out all good

jaimejiyepark:
After this pr goes through, ill put in a pr for cardverify since it uses cardscanbundle once

kingst:
For this change let's just change CardScan to CSBundle and leave the rest of the names as is. Otherwise, LGTM.

PR #142, opened 2020-03-24
recreating group set up

jaimejiyepark:
* made navigation controller since it looks like their payment page is a nav controller from top bard
* made nav controller full screen like the images
* formsheet for just cardscan vc

### DIFF
--- a/Example/CardScan/Base.lproj/Main.storyboard
+++ b/Example/CardScan/Base.lproj/Main.storyboard
@@ -261,7 +261,7 @@
         <!--Modal View View Controller-->
         <scene sceneID="KW8-5r-Gdc">
             <objects>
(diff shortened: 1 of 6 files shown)


jaimejiyepark:
* ```ScanBase.SupportInterface...``` is needed for when parent nav vc doesn't take scan vc's supported interface orientations
* ScanVc's ```shouldAutoRotate`` is true for when parent does take scan vc's supported interfaces orientations

Example/CardScan/Base.lproj/Main.storyboard
```diff
@@ -20,7 +20,7 @@
                         <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
```
jaimejiyepark: storyboard changes: added two buttons ```present modal taking child's orientations``` and ```present modal without taking child's orientations```

PR #147, opened 2020-04-16
add view obscuring during background 

jaimejiyepark:
Added view obscuring per Doyensec's suggestion: 
```To protect the iOS applications, implement a similar minor mitigation. Some of these include hiding the sensitive informations by using UIScreen you to detect when the screen is being captured or when the user move to the background the application```

Decided to add another pair of observers for when the application goes into background / foreground since delegate-communication won't work with parent-child relationship between `scanBaseViewController` and `scanViewController`. A few other reasons is that the notification handler in `scanBaseViewController` is used specifically for the `machineLearningQueue` and the view is only accessible in the `scanViewController` level.

### DIFF
--- a/CardScan/Classes/ScanViewController.swift
+++ b/CardScan/Classes/ScanViewController.swift
@@ -297,6 +297,11 @@ public protocol MainLoopDelegate: class {
         self.cornerView.layer.borderColor = self.cornerBorderColor
     }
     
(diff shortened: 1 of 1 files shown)


jaimejiyepark:
yea i think so. let me add that

jaimejiyepark:
@kingst  i noticed that the notification observers in scanbase arent ever removed. do you want me to remove them?

PR #148, opened 2020-04-21
migrating to swift5

jaimejiyepark:
* only major change was to appDelegate but since it's example-side, I didn't add backwards compatibility.
* We don't use any deprecated methods that are changed listed in the Swift5 docs.

### DIFF
--- a/CardScan.podspec
+++ b/CardScan.podspec
@@ -9,7 +9,8 @@ CardScan is a library for scanning credit cards.
   s.homepage         = 'https://cardscan.io'
   s.license          = { :type => 'BSD', :file => 'LICENSE' }
   s.author           = { 'Sam King' => 'kingst@gmail.com' }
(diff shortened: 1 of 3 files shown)


jaimejiyepark:
Also ran all the tests just in case we catch anything funky and we're clean there too

CardScan.podspec
```diff
@@ -9,7 +9,8 @@ CardScan is a library for scanning credit cards.
   s.homepage         = 'https://cardscan.io'
   s.license          = { :type => 'BSD', :file => 'LICENSE' }
   s.author           = { 'Sam King' => 'kingst@gmail.com' }
-  s.source           = { :git => 'https://github.com/Ibotta/cardscan-ios.git', :tag => s.version.to_s }
+  s.source           = { :git => 'https://github.com/getbouncer/cardscan-ios.git', :tag => s.version.to_s }
```
justinvallely, line 12: Whoops. My fault! I should have excluded that when comparing against our fork. So sorry.

PR #151, opened 2020-04-26
landscape support

jaimejiyepark:
* PreviewView and VideoFeed takes interface orientation
* Added VC & Nav like Groupons to MainStoryboard


jaimejiyepark:
Hey Sam, just saw this. 
I tested it on my iPad and it worked fine? I'll send you a screenshot through Slack.

PR #153, opened 2020-04-30
Producer consumer ocr

kingst:
Introduces the consumer / producer OCR. Highlights include:
- Significantly faster frame rate
- Name extraction
- Better expiry


jaimejiyepark:
I tested it out with my phone and wow the legal name extraction is really cool

CardScan/Classes/CreditCardOcr/CreditCardOcrImplementation.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    init(dispatchQueue: DispatchQueue) {
+        self.dispatchQueue = dispatchQueue
+    }
+    
+    func recognizeCard(in fullImage: CGImage, roiRectangle: CGRect) -> CreditCardOcrPrediction {
+        preconditionFailure("This method must be overridden")
+    }
+    
+    func croppedImage(fullCardImage: CGImage, roiRectangle: CGRect) -> CGImage? {
```
jaimejiyepark, line 37: would it make sense to put these cropping functions in imageUtils?

CardScan/Classes/CreditCardOcr/AppleCreditCardOcr.swift
```diff
@@ -0,0 +1,78 @@
...
+        semaphore.wait()
+        let duration = -startTime.timeIntervalSinceNow
+        self.computationTime += duration
+        self.frames += 1
+
+        return CreditCardOcrPrediction(image: image, number: pan, expiryMonth: expiryMonth, expiryYear: expiryYear, name: name, computationTime: duration, numberBoxes: numberBox.map { [$0] }, expiryBoxes: expiryBox.map { [$0] }, nameBoxes: nameBox.map { [$0] })
+    }
+    
+    static func likelyName(_ text: String) -> String? {
+        let lettersAndSpace = text.reduce(true) { acc, value in
```
jaimejiyepark, line 61: what's acc?

CardScan/Classes/CreditCardOcr/CreditCardOcrPrediction.swift
```diff
@@ -0,0 +1,74 @@
...
+            return nil
+        }
+
+        let result = regex.matches(in: string, range: NSRange(string.startIndex..., in: string))
+        
+        if result.count == 0 {
+            return nil
+        }
+        
+        let range1 = Range(result.first!.range(at: 1), in: string)!
```
jaimejiyepark, line 60: I know `result.first` is taken care of by the previous `result.count` conditional statement but for the range would it be safer to use a default value? Even though we know result will have something but the `!` gives me nightmares.

I'm not sure though.

CardScan/Classes/CreditCardOcr/CreditCardOcrImplementation.swift
kingst: yeah it does, but let's do that later. There is still a ton of refactoring that we need to do.

CardScan/Classes/CreditCardOcr/AppleCreditCardOcr.swift
kingst: accumulator

CardScan/Classes/CreditCardOcr/CreditCardOcrPrediction.swift
kingst: yikes! I'll fix that, nice catch.

PR #157, opened 2020-05-21
Resume threads if deallocated when in background

kingst:
I wasn't able to trigger this bug myself, but I did see a report from Crashlytics on the user study and I'm 90% sure that this fixes it.

### DIFF
--- a/CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
+++ b/CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
@@ -249,7 +249,7 @@ public class OcrMainLoop : MachineLearningLoop {
     //   * The willResignActive function blocks the transition to the background until
     //     it completes, which we couldn't find docs on but verified experimentally
     @objc func willResignActive() {
-        inBackground = true
+        mutexQueue.sync { self.inBackground = true }
         // this makes sure that any currently running predictions finish before we
         // let the app go into the background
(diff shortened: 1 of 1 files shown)


jaimejiyepark:
shouldnt the mlqueue resuming and suspension be in the mutexqueue as well to ensure they behave properly before it backgrounds or not?

CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
```diff
@@ -260,7 +260,7 @@ public class OcrMainLoop : MachineLearningLoop {
     }
     
     @objc func didBecomeActive() {
-        inBackground = false
+        mutexQueue.sync { self.inBackground = false }
         for queue in machineLearningQueues {
```
jaimejiyepark, line 264: should this be in the mutexQueue?

kingst: it's just to make sure that all shared state is accessed from the `mutexQueue` note that we're using `sync` here

kingst:
The sync call on the machine learning queues make sure that the ML is done before the thread is suspended

jaimejiyepark:
but the for loop isnt in the mutex queue. shouldnt it be in the mutex queue?

jaimejiyepark:
#shipit

PR #158, opened 2020-05-22
Updating name rules

jaimejiyepark:
name rule changes:
* get rid of non-name words & numbers
* if remaining words (in order) have a count of 2 or 3, then it's predicted name

motivations:
* name words can clump with non-name words in detection


Example/CardScan/Base.lproj/Main.storyboard
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="15505" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="vXZ-lx-hvc">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="16097" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="vXZ-lx-hvc">
```
jaimejiyepark, line 2: main.storyboard changes: change text color from default -> dark to be able to see the results

jaimejiyepark:
(1) I can extend TestOcr but it'll take a little bit of time.
(2) Carthage builds!

kingst:
Let's not block on TestOcr as long as we have a set of cards that we test manually on I'm fine shipping this.

PR #159, opened 2020-05-25
Rewrote full screen and ROI image cropping logic 

jaimejiyepark:


jaimejiyepark:
This should work for all portrait cases. 

Next PR should work for all cases, specifically rotation cases

CardScan/Classes/Extensions/Image+utils.swift
```diff
@@ -35,51 +35,44 @@ extension CGImage {
...
+        let fullScreenImage: CGImage? = {
+            // if image is already 16:9, no need to crop to match crop ratio
+            if cropRatio == imageAspectRatio {
+                return self
+            } else {
+                // imageAspectRatio not being 16:9 implies image being in landscape
+                // get width to first not cut out any card information
+                let cropWidth = previewViewFrame.width * pointsToPixel
+                // then get height to match 16:9 ratio
+                let cropHeight = cropWidth * (16.0 / 9.0)
```
kingst, line 58: There is a corner case here where the cropHeight might be too big for the image that we have that we can handle. If you're confident that we won't see it in practice yet, I suggest adding a `// todo(jaime): handle case where height is too big for the image` comment to track it and tackle it with the next pr.

PR #163, opened 2020-06-05
Orientation rotation support

jaimejiyepark:
* Every view should be portrait **_regardless_**  of how you enter `scanViewController`(except iPad with formsheet presentation)

To Test: 
* press `Present Modally`
* change presentationStyle in `modalScanViewController` between `[.formsheet, .pagesheet]`
* press `toScanViewController` or `to modal ScanViewController` while in landscape or portrait


kingst:
Oh I almost forgot one optional item. The debug window shows the card shrunk by a bit because we crop at 10% beyond the ROI rectangle. Not sure it's worth fixing, but FYI

PR #167, opened 2020-06-11
Replace legacy with DD Ocr

dxaen:
The PR replaces our legacy model with the new OCR model. Some highlighting features of the new model:

- We add iPhone5 and above to our supported device list.

- The overall model size is down to 1.7 MB from 3.1 MB.

### TestOcr results

| Device  | Precision | Recall | Avg. Frame Time |
| ------------- | ------------- | ------------- | -------------|
| iPhone5s  | 99.4%  | 55.9% | 0.552s |
| iPadMini2  | 99.4%  | 55.9% | 0.534s |


CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
```diff
@@ -77,6 +77,7 @@ public class OcrMainLoop : MachineLearningLoop {
     public enum AnalyzerType {
         case apple
         case legacy
```
kingst, line 79: Remove -- we don't have legacy anymore

CardScan/Classes/CreditCardOcr/SSDCreditCardOcr.swift
```diff
@@ -0,0 +1,28 @@
...
+        guard let image = croppedImageForSsd(fullCardImage: fullImage, roiRectangle: roiRectangle)
+            else {
+                return CreditCardOcrPrediction.emptyPrediction(cgImage: fullImage)
+        }
+                
+        let ocr = OcrDD()
+        let startTime = Date()
+        let number = ocr.perform(croppedCardImage: image)
+        let duration = -startTime.timeIntervalSinceNow
+        print("DEBUG: Inference time for DD Ocr: ", duration)
```
kingst, line 21: let's remove the debugging print statements or have a flag that defaults to "off" for them

CardScan/Classes/DetectedSSDOcrBox.swift
```diff
@@ -0,0 +1,32 @@
...
+        
+        self.label = category
+        self.confidence = conf
+        self.rect = CGRect(x: XMin_, y: YMin_, width: XMax_ - XMin_, height: YMax_ - YMin_)
+        self.imgSize = imageSize
+    }
+    
+     func toDict() -> [String: Any] {
+        
+        return ["":""]
```
kingst, line 30: We'll want to check with the Card Verify flow to make sure that this model returns whatever our server-side is expecting for fraud checks.

CardScan/Classes/GeneratedSSD.swift
```diff
@@ -0,0 +1,184 @@
...
+// SSDOcr.swift
+//
+// This file was automatically generated and should not be edited.
+//
+
+import CoreML
+
+
+/// Model Prediction Input Type
+@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
```
kingst, line 12: I know that this file was auto-generated but all of the forced unwraps make me really really nervous. Can we pick those out by hand? We might need to change the high level interface a bit for it but I think it's worth it.

CardScan/Classes/PredictionUtilOcr.swift
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * A utitliy struct that applies non-max supression to each class
+     * picks out the remaining boxes, the class probabilities for classes
+     * that are kept and composes all the information in one place to be returned as
+     * an object.
+     */
+    func predictionUtil(scores: [[Float]], boxes: [[Float]], probThreshold: Float,
+                       iouThreshold: Float, candidateSize: Int , topK: Int) -> Result{
+        var pickedBoxes:[[Float]] = [[Float]]()
+        var pickedLabels:[Int] = [Int]()
```
kingst, line 21: nit: since you have the type you can initialize it using `[]`

CardScan/Classes/SSDOcrDetect.swift
```diff
@@ -0,0 +1,274 @@
...
+                                                               XMin: Double(result.pickedBoxes[idx][0]), YMin: Double(result.pickedBoxes[idx][1]),
+                                                               XMax: Double(result.pickedBoxes[idx][2]), YMax: Double(result.pickedBoxes[idx][3]),
+                                                               imageSize: image.size))
+        }
+        
+        if self.isQuickRead(allBoxes: DetectedOcrBoxes){
+            let _cardNumber = processQuickRead(allBoxes: DetectedOcrBoxes)
+            return _cardNumber
+        }
+        else {
```
kingst, line 133: remove `else` since it doesn't do anything

CardScan/Classes/SSDOcrDetect.swift
```diff
@@ -0,0 +1,274 @@
...
+            //os_log("%@", type: .error, "Not Quick Read")
+        }
+        
+        if (!result.pickedBoxes.isEmpty) {
+                let topCordinates = result.pickedBoxes.map{$0[1]}
```
kingst, line 138: hard coded array indexes are risky in swift since we can't catch them if we go OOB. Can we add some unit tests here to make sure that it is safe?

CardScan/Classes/SSDOcrDetect.swift
```diff
@@ -0,0 +1,274 @@
...
+      
+        for idx in 0..<fourthGroup.count {
+            _cardNumber = _cardNumber + String(fourthGroup[idx].label)
+        }
+        
+        if CreditCardUtils.isValidNumber(cardNumber: _cardNumber){
+            print(_cardNumber)
+            return _cardNumber
+        }
+        else {
```
kingst, line 214: remove

CardScan/Classes/SSDOcrDetect.swift
```diff
@@ -0,0 +1,274 @@
...
+            //os_log("%@" , type: .debug, "Could verify \(_cardNumber)")
+        }
+        return nil
+        
+    }
+    
+    func isQuickRead(allBoxes: DetectedAllOcrBoxes) -> Bool {
+        if (allBoxes.allBoxes.isEmpty) || (allBoxes.allBoxes.count != 16) {
+            //os_log("%@", type: .debug, "Failed in capacity:\(allBoxes.allBoxes.capacity)")
```
kingst, line 223: remove

CardScan/Classes/SSDOcrOutputExtensions.swift
```diff
@@ -0,0 +1,141 @@
+//
+//  SSDOcrOutputExtensions.swift
+//  CardScan
+//
+//  Created by xaen on 3/22/20.
+//
+
+import Foundation
+import Accelerate
```
kingst, line 9: what is this library? Is it a standard iOS library? We probably need to add something to the podspec file to make sure that the app always includes it

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -132,8 +132,11 @@ public protocol TestingImageDataSource: AnyObject {
                 return true
             }
             switch Api.deviceType() {
-            case "iPhone3,1", "iPhone3,2", "iPhone3,3", "iPhone4,1", "iPhone5,1", "iPhone5,2", "iPhone5,3", "iPhone5,4", "iPhone6,1", "iPhone6,2", "iPhone7,2", "iPhone7,1":
+            //only block iphone 4 and below
```
kingst, line 135: with this change we should be able to remove all device specific checks since the OS check will suffice. I suggest keeping this function but just have it always return true.

CardScan/Classes/SSDOcrOutputExtensions.swift
dxaen: This is a standard library that uses CPU's vector instructions just like Metal uses the GPU.
This is already in the SDK and we have been using this since August'19.
https://github.com/getbouncer/cardscan-ios/blob/master/CardScan/Classes/SSDOutputExtensions.swift

CardScan/Classes/SSDOcrDetect.swift
dxaen: How about creating an extension to Arrays that checks if this index exists and returns nil otherwise?

PR #171, opened 2020-06-16
Liveness ml

kingst:
This consists mostly of exposing classes and variables so that we can use them in our liveness check.

### DIFF
--- a/CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
+++ b/CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
@@ -82,7 +82,7 @@ open class OcrMainLoop : MachineLearningLoop {
     public var scanStats = ScanStats()
     
     public weak var mainLoopDelegate: OcrMainLoopDelegate?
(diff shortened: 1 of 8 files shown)


CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
```diff
@@ -92,21 +92,30 @@ public class OcrMainLoop : MachineLearningLoop {
...
             }
-            machineLearningQueues.append(queue)
+        }
+        setupMl(ocrImplementations: ocrImplementations)
+    }
+    
+    /// Note: you must call this function in your constructor
+    public func setupMl(ocrImplementations: [CreditCardOcrImplementation]) {
+        machineLearningQueues = []
+        scanStats.model = "legacy+apple"
```
jaimejiyepark, line 115: should the scan stats model be called legacy+apple if some devices can support legacy but not apple?

kingst: the system is still "legacy+apple" but not all devices get apple

PR #173, opened 2020-06-17
Refactor for the completion loop

kingst:
This PR consists mostly of cleaning up variable names and exposing UX model results. There is also a bit of dead code removal.


jaimejiyepark:
#shipit

PR #175, opened 2020-06-18
Fixing rotation in release build config

jaimejiyepark:
Preview view video orientation is set after video set up since connection is nil until video feed set up is complete

### DIFF
--- a/CardScan/Classes/ScanBaseViewController.swift
+++ b/CardScan/Classes/ScanBaseViewController.swift
@@ -217,11 +217,12 @@ public protocol TestingImageDataSource: AnyObject {
         self.ocrMainLoop()?.mainLoopDelegate = self
         self.previewView?.videoPreviewLayer.session = self.videoFeed.session
         
-        self.videoFeed.videoOrientation = self.initialVideoOrientation
         self.videoFeed.pauseSession()
         //Apple example app sets up in viewDidLoad: https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/avcam_building_a_camera_app
-        self.videoFeed.setup(captureDelegate: self, completion: { success in
-            self.previewView?.videoPreviewLayer.connection?.videoOrientation = self.initialVideoOrientation
+        self.videoFeed.setup(captureDelegate: self, initialVideoOrientation: self.initialVideoOrientation, completion: { success in
+            if self.previewView?.videoPreviewLayer.connection?.isVideoOrientationSupported ?? false {
+                self.previewView?.videoPreviewLayer.connection?.videoOrientation = self.initialVideoOrientation
+            }
             if let level = torchLevel {
                 self.setTorchLevel(level: level)
             }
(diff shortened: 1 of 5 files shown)


CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -184,44 +191,37 @@ public protocol TestingImageDataSource: AnyObject {
...
+        }
         
         if testingImageDataSource != nil {
             self.ocrMainLoop()?.imageQueueSize = 20
         }
         
+        self.ocrMainLoop()?.mainLoopDelegate = self
+        self.previewView?.videoPreviewLayer.session = self.videoFeed.session
+        
+        self.videoFeed.videoOrientation = self.initialVideoOrientation
```
kingst, line 220: I suggest getting rid of the `videoOrientation` variable in `videoFeed` and just pass the initial orientation into the setup function -- I think that it only gets used there.

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -184,44 +191,37 @@ public protocol TestingImageDataSource: AnyObject {
...
         self.videoFeed.pauseSession()
         //Apple example app sets up in viewDidLoad: https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/avcam_building_a_camera_app
         self.videoFeed.setup(captureDelegate: self, completion: { success in
+            self.previewView?.videoPreviewLayer.connection?.videoOrientation = self.initialVideoOrientation
```
kingst, line 224: you should check if orientation setting is supported before setting it: https://developer.apple.com/documentation/avfoundation/avcaptureconnection/1389415-videoorientation

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -184,44 +191,37 @@ public protocol TestingImageDataSource: AnyObject {
...
-    }
-    
     public func setupOnViewDidLoad(regionOfInterestLabel: UIView, blurView: BlurView, previewView: PreviewView, cornerView: CornerView?, debugImageView: UIImageView?, torchLevel: Float?) {
         
         self.regionOfInterestLabel = regionOfInterestLabel
         self.blurView = blurView
         self.previewView = previewView
         self.debugImageView = debugImageView
         self.cornerView = cornerView
+        ScanBaseViewController.isPadAndFormsheet = UIDevice.current.userInterfaceIdiom == .pad && self.modalPresentationStyle == .formSheet
```
kingst, line 201: can we make `isPadAndFormsheet` a member variable instead of a static? You can pass it into the setup function for the videoFeed if needed, but having this as a static can be problematic because it's specific to an individual ViewController and not something that is global

CardScan/Classes/ScanBaseViewController.swift
kingst: Also you should probably only do this on the success case? We need to test with permissions to make sure that it's working.

CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: i made it static to use in the static supportedInterfaces function

kingst: ugg ok ignore this for now

jaimejiyepark: :-|

CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: okiedokie changed. also i deleted the app and ran it again to check with the initial permissions

PR #179, opened 2020-06-23
Add the bounding box of individual digits to the items the OCR returns.

dxaen:


CardScan/Classes/OcrDD.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
 
 
 public class OcrDD{
+    public var scanStats = ScanStats()
```
kingst, line 12: instead of using scanstats for this why not create a variable that's just for a list of boxes?

dxaen: This is consistent with how Ocr.swift does it, incase we want to add more stuff to this. But I can create a variable specific to this if you prefer.

kingst: Yeah, OCR using scanstats is a bit of a throwback to the old way of doing it -- we track the real scan stats in the main loop class. Let's go with a specific variable.

dxaen: Alright.

CardScan/Classes/DetectedAllOcrBoxes.swift
```diff
@@ -15,4 +15,8 @@ public struct DetectedAllOcrBoxes {
         let frameArray = self.allBoxes.map{$0.toDict()}
         return frameArray
     }
+    
+    public func getBoundingBoxesOfDigits() -> [CGRect] {
+        return self.allBoxes.map{$0.rect}
```
jaimejiyepark, line 20: nit: need spaces. `self.allBoxes.map { $0.rect }`

CardScan/Classes/OcrDD.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
```
jaimejiyepark, line 10: nit: get rid of one new line

CardScan/Classes/OcrDD.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
 
 
 public class OcrDD{
+    public var lastDetectedBoxes: [CGRect] = []
     public init(){}
```
jaimejiyepark, line 13: do we need this init if it's not used?

jaimejiyepark:
sorry about all the nits! besides them, looks good! #shipit

CardScan/Classes/DetectedAllOcrBoxes.swift
dxaen: yes ma'am.

CardScan/Classes/OcrDD.swift
dxaen: 👍

CardScan/Classes/OcrDD.swift
dxaen: I need to add some code here as in Ocr.swift. Thanks for pointing this out.

PR #180, opened 2020-07-04
Sets the default for the iPad to a formSheet

kingst:
On the iPad if you don't select "requires full screen" in the `Info.plist` file then it won't use the standard mechanisms for locking the screen orientation and our full screen viewcontroller won't work.

This patch sets the presentation style to `formSheet` by default for iPads and adds a comment to the code + updates the README to let people know.


jaimejiyepark:
Pressed `Require full screen` but the scanViewController autos it to formsheet, so how can a user utilize the full screen?

kingst:
They can just set it on the viewcontroller themselves by setting `.modalPresentationStyle`

jaimejiyepark:
#shipit

PR #181, opened 2020-07-08
Remove the old legacy OCR code

kingst:
Removes the legacy OCR code and get rid of the static model for SSDOcr so that we can make it thread safe if needed.


jaimejiyepark:
#shipit

PR #183, opened 2020-07-09
Ocr main loop deadlock bug fix

kingst:
Previously by having some of our queue clean up logic live in the `deinit` function of our main loop, it got hard to reason about because objects and deinit in subtle ways (e.g., see the first commit of this PR to see the bug). This PR cleans up our backgrounding logic, simplifies our use of dispatch queues, and removes the logic we had on `deinit` previously.

This new backgrounding logic is much easier to reason about, so hopefully this one will stick.


jaimejiyepark:
i cant seem to reproduce the bug

jaimejiyepark:
lgtm

PR #184, opened 2020-07-13
Add navigation bar configuration

jaimejiyepark:
Logic changes made in `ScanBase` and `ScanView`

- changed main.storyboard to add navigation bar + bar functionality
- moved delegate logic to extension like AJ mentioned


jaimejiyepark:
I don't see any difference between it being animated or not. Should i add it in anyways?

jaimejiyepark:
and yes, if not set then its hidden as before

PR #187, opened 2020-07-23
Fix bug with Xcode 12 beta 3 compiler optimizations

kingst:
When the clang code generator and swift compiler use optimizations in Xcode 12 Beta 3 CardScan doesn't compile. To reproduce:

- Download Xcode 12 Beta 3
- Choose "build for any device"
- Run an "Archive"

### DIFF
--- a/CardScan/Classes/Expiry.swift
+++ b/CardScan/Classes/Expiry.swift
@@ -1,9 +1,15 @@
 import Foundation
 
-public struct Expiry: Hashable {
+public class Expiry: Hashable {
     public let string: String
     public let month: UInt
     public let year: UInt
+    
+    init(string: String, month: UInt, year: UInt) {
+        self.string = string
+        self.month = month
+        self.year = year
+    }
     
     public static func == (lhs: Expiry, rhs: Expiry) -> Bool {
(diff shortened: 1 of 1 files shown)


jaimejiyepark:
I get another error with the changes?

kingst:
Make sure that you select the beta version of xcode as your default using:

`sudo xcode-select -s /Applications/Xcode-beta.app`

kingst:
Actually, I'm going to hold off on committing this for a bit -- Jaime is right, there are a bunch of other places that are impacted by this compiler bug. We're going to wait until late August and if the problem persists we'll add a work around well in advance of the September release of Xcode 12 and iOS 14.

kingst:
I just tested the repo out with Xcode 12 beta4 and it seems like Apple fixed their compiler bug. @jaimejiyepark can you confirm? If so I'll close this PR.

jaimejiyepark:
It hasn't been fixed on my end.

kingst:
@justinvallely I'm going to close this for now, let us know after you test out Xcode 12 beta 4 to confirm that we're good on your end as well. If not we'll re-open this.

PR #188, opened 2020-07-25
Add a SimpleScanViewController for full UI customization

kingst:
This PR is for our programmatic UI extension to `ScanBaseViewController`. We'll use this to show how people can fully customize our UI. This new class is meant to be complete but bare bones from a UI perspective.  People can use it directly if they like the look and feel, or the can modify variables or subclass for more control.

In a future PR we'll:
- Add examples to show simple UI customizations.
- Migrate our fraud ViewControllers over to use this.
- Migrate `ScanViewController` over to use this as a base class.


jaimejiyepark:
looks good. one thing though: is it a good idea to have another scan delegate protocol for simple that looks exactly like the original? or should we make the scanVC protocol more generic to accommodate simple

kingst:
It's a good question and my first cut did this but I created a separate protocol for two reasons (1) I don't want to change public interfaces unless we get a massive simplification, which isn't there case here and (2) we're going to use this as the base for fraud challenges which don't need skip, so the interface isn't a good match

jaimejiyepark:
hmm alright then. something to keep in mind. #shipit

PR #189, opened 2020-07-30
Adding model config downloading to API

jaimejiyepark:
Manually tested with 
```swift 
Api.apiKey = "<api-key>"
    Api.getModelDownloadConfig(endpoint: "/v1/model/ios/bob/1", parameters: [:]) { response, error in
        guard let response = response, error == nil else {
            print("oh no")
            return
         }
     print(response)
}
```


CardScan/Classes/API.swift
```diff
@@ -135,6 +135,61 @@ public struct Api {
...
+        modelDownloadConfig(endpoint: endpoint, parameters: parameters, completion: completion)
+    }
+    
+    private static func modelDownloadConfig(endpoint: String, parameters: [String: Any], completion: @escaping ApiCompletion) {
+        guard let baseUrl = self.baseUrl else {
+            DispatchQueue.main.async { completion(nil, apiUrlNotSet) }
+            return
+        }
+
+        guard let url = URL(string: baseUrl + endpoint) else {
```
kingst, line 153: Can you add a function that is something like `urlWithQueryParameters(baseUrl, endpoint, parameters) -> URL?` so that we can unit test it? I saw from SO they suggest escaping `+` manually, but this is the sort of thing that we'd want to have a unit test for so putting this bit of code into its own function will make testing it easy

CardScan/Classes/API.swift
```diff
@@ -190,6 +178,14 @@ public struct Api {
         }.resume()
     }
     
+    static func urlWithQueryParameters(baseUrl: String, endpoint: String, parameters: [String: Any]) -> URL? {
+        var components = URLComponents(string: baseUrl + endpoint)
+        components?.queryItems = parameters.map { (key, value) in URLQueryItem(name: key, value: value as? String) }
```
kingst, line 183: Sorry, I should have caught this on my first read of the PR.

If you get a parameter that isn't a string it will silently fail. I suggest making `parameters` of type `[String: String]` or try to handle cases and error out if you get one that you aren't expecting instead of setting the value to nil

jaimejiyepark:
@kingst  added `modelDownloadConfigTest` and made changes in `API`

CardScan/Classes/API.swift
```diff
@@ -135,6 +135,67 @@ public struct Api {
...
+                }
+            }
+        }.resume()
+    }
+    
+    static func urlWithQueryParameters(baseUrl: String, endpoint: String, parameters: [String: Any]) -> URL? {
+        let parametersDict: [String: String] = {
+            var parametersDict: [String: String] = [:]
+            for (key, value) in parameters {
+                if let val = value as? String {
```
kingst, line 185: I think what you want here is to replace this whole `if` block with:

`let val = "\(value)"`

The current implementation still silently ignores non string values, which I think we should avoid. My suggestion will at least convert everything to a string, but I can also see the case for erroring out here and requiring the caller to set it.

CardScan/Classes/API.swift
```diff
@@ -135,6 +135,67 @@ public struct Api {
...
+                    parametersDict[key] = val
+                }
+            }
+            return parametersDict
+        }()
```
kingst, line 190: I don't think that you need to encapsulate this in a block, just do something like this:

```
var stringParameters: [String: String] = [:]
for (key, value) in parameters {
    stringParameters[key] = "\(value)"
}
```

CardScan/Classes/API.swift
kingst: Also, even with this I'd still be careful to check for arrays / dictionaries / optionals  and error out if you get something that you're not expecting.

CardScan/Classes/API.swift
```diff
@@ -135,6 +135,62 @@ public struct Api {
...
+                if "error" == responseData["status"] as? String {
+                    completion(nil, ApiError(response: responseData))
+                } else {
+                    completion(responseData, nil)
+                }
+            }
+        }.resume()
+    }
+    
+    static func urlWithQueryParameters(baseUrl: String, endpoint: String, parameters: [String: Any]) -> URL? {
```
kingst, line 181: I'd add a comment that says something like "we are going to cast all of the parameter values to strings so make sure that you don't have any arrays, dicts, optionals, or other values that won't convert cleanly to strings"

PR #192, opened 2020-08-01
fix unit tests

dxaen:
This commit fixes the unit tests. This was needed since `ssdOcr` model instance inside `SSDOcrDetect` is no longer a static variable.


dxaen:
for some reason it is failing the check in GitHub.

kingst:
> for some reason it is failing the check in GitHub.

Looks like this is your error:

```Build input file cannot be found: '/Users/runner/work/cardscan-ios/cardscan-ios/Example/CardScan_ExampleTests/CardScan_BundleTests.swift'
```

Not sure if this is caused by your PR or something else, but hopefully @awushensky can help out

kingst:
It looks like you deleted this file as a part of your PR and I just tested the build on master, which works, so it is this PR. Here's my suggestion: run `pod install` in the Example directory and include the changes from that in your PR. It looks like you're deleting the file from the file system, but my guess is that you haven't updated the project file, thus the mismatch. Also, if you want to try running the test the same way that it does in CI you can run this command:

```bash
cd Example
pod install
xcodebuild test -workspace CardScan.xcworkspace -scheme CardScan_Example -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.6'
```

jaimejiyepark:
@dxaen  go to targets -> cardscan_exampletest -> build phase -> delete the bundle test giving you issues.

dxaen:
> It looks like you deleted this file as a part of your PR and I just tested the build on master, which works, so it is this PR. Here's my suggestion: run `pod install` in the Example directory and include the changes from that in your PR. It looks like you're deleting the file from the file system, but my guess is that you haven't updated the project file, thus the mismatch. Also, if you want to try running the test the same way that it does in CI you can run this command:
> 
> ```shell
> cd Example
> pod install
> xcodebuild test -workspace CardScan.xcworkspace -scheme CardScan_Example -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.6'
> ```

If I include those files, my apple dev id gets added to the diff. Is there a way I can prevent that?

kingst:
> > It looks like you deleted this file as a part of your PR and I just tested the build on master, which works, so it is this PR. Here's my suggestion: run `pod install` in the Example directory and include the changes from that in your PR. It looks like you're deleting the file from the file system, but my guess is that you haven't updated the project file, thus the mismatch. Also, if you want to try running the test the same way that it does in CI you can run this command:
> > ```shell
> > cd Example
> > pod install
> > xcodebuild test -workspace CardScan.xcworkspace -scheme CardScan_Example -destination 'platform=iOS Simulator,name=iPhone 11,OS=13.6'
> > ```
> 
> If I include those files, my apple dev id gets added to the diff. Is there a way I can prevent that? @kingst

I just updated the Growth Metrics account so that you can manage certificates -- can you try to do it using the Growth Metrics org?

PR #195, opened 2020-08-03
added model download func

jaimejiyepark:
Added function to download models with signed url


CardScan/Classes/API.swift
```diff
@@ -135,6 +135,23 @@ public struct Api {
         apiCall(endpoint: endpoint, parameters: apiParameters, completion: completion)
     }
     
+    static func modelDownload(signedUrl: String, completion: @escaping ApiCompletion) {
+        guard let url = URL(string: signedUrl) else {
+            completion(nil, apiUrlNotSet)
```
kingst, line 140: All of our other API calls post the `completion` routine on the main queue, we should be consistent about this.

PR #196, opened 2020-08-04
Support dot expiry

awushensky:


awushensky:
I've done some manual testing for right now using our data sets. I did not encounter any false positives.

PR #197, opened 2020-08-05
Main loop state machine

kingst:
This is a simple state machine for OCR -- we'll build off of this for the more complex fraud flows in a later PR.

### DIFF
--- a/CardScan/Classes/CreditCardOcr/CreditCardOcrResult.swift
+++ b/CardScan/Classes/CreditCardOcr/CreditCardOcrResult.swift
@@ -13,17 +13,19 @@ public class CreditCardOcrResult: MachineLearningResult {
     public let number: String
     public let expiry: String?
     public let name: String?
+    public let state: MainLoopState
     public let isFinished: Bool
     
     // this is only used by Card Verify and the Liveness check and filled in by the UxModel
     public var hasCenteredCard: CenteredCardState?
     
-    init(mostRecentPrediction: CreditCardOcrPrediction, number: String, expiry: String?, name: String?, isFinished: Bool, duration: Double, frames: Int) {
+    init(mostRecentPrediction: CreditCardOcrPrediction, number: String, expiry: String?, name: String?, state: MainLoopState, duration: Double, frames: Int) {
         self.mostRecentPrediction = mostRecentPrediction
         self.number = number
         self.expiry = expiry
         self.name = name
-        self.isFinished = isFinished
+        self.state = state
+        self.isFinished = state == .finished
         super.init(duration: duration, frames: frames)
     }
     
@@ -35,7 +37,7 @@ public class CreditCardOcrResult: MachineLearningResult {
     }
     
     public static func finishedWithNonNumberSideCard(prediction: CreditCardOcrPrediction, duration: Double, frames: Int) -> CreditCardOcrResult {
-        let result = CreditCardOcrResult(mostRecentPrediction: prediction, number: "", expiry: nil, name: nil, isFinished: true, duration: duration, frames: frames)
+        let result = CreditCardOcrResult(mostRecentPrediction: prediction, number: "", expiry: nil, name: nil, state: .finished, duration: duration, frames: frames)
         result.hasCenteredCard = .nonNumberSide
         return result
     }
--- a/CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
+++ b/CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
@@ -1,0 +1,62 @@
+//
+//  MainLoopStateMachine.swift
+//  CardScan
+//
+//  Created by Sam King on 8/5/20.
+//
+
+import Foundation
+
+public enum MainLoopState {
+    case initial
+    case ocrOnly
+    case cardOnly
+    case ocrAndCard
+    case finished
+}
+
+public protocol MainLoopStateMachine {
+    func loopState() -> MainLoopState
+    func event(prediction: CreditCardOcrPrediction) -> MainLoopState
+}
+
+open class OcrMainLoopStateMachine: MainLoopStateMachine {
+    
+    var state: MainLoopState = .initial
+    var startTimeForCurrentState = Date()
+    var timeOfFirstOcr: Date?
+    
+    public let errorCorrectionDuration = 2.0
+    
+    public func loopState() -> MainLoopState {
+        return state
+    }
+    
+    public func event(prediction: CreditCardOcrPrediction) -> MainLoopState {
+        if prediction.number != nil && timeOfFirstOcr == nil {
+            timeOfFirstOcr = Date()
+        }
+                
+        let currentState = state
+        state = transition(prediction: prediction)
+        if state != currentState {
+            startTimeForCurrentState = Date()
+        }
+        
+        return state
+    }
+    
+    open func transition(prediction: CreditCardOcrPrediction) -> MainLoopState {
+        let timeInCurrentState = -startTimeForCurrentState.timeIntervalSinceNow
+        
+        switch (state, timeInCurrentState, timeOfFirstOcr) {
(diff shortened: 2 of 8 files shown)


CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,62 @@
...
+    func event(prediction: CreditCardOcrPrediction) -> MainLoopState
+}
+
+open class OcrMainLoopStateMachine: MainLoopStateMachine {
+    
+    var state: MainLoopState = .initial
+    var startTimeForCurrentState = Date()
+    var timeOfFirstOcr: Date?
+    
+    public let errorCorrectionDuration = 2.0
```
awushensky, line 29: nit: Name this with the units, so it's clear that this is seconds and not milliseconds.

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,62 @@
...
+        state = transition(prediction: prediction)
+        if state != currentState {
+            startTimeForCurrentState = Date()
+        }
+        
+        return state
+    }
+    
+    open func transition(prediction: CreditCardOcrPrediction) -> MainLoopState {
+        let timeInCurrentState = -startTimeForCurrentState.timeIntervalSinceNow
```
awushensky, line 50: nit: name this with units

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,62 @@
...
+        
+        switch (state, timeInCurrentState, timeOfFirstOcr) {
+        case (.initial, _, .some):
+            return .ocrOnly
+        case (.ocrOnly, errorCorrectionDuration..., _):
+            return .finished
+        default:
```
awushensky, line 53: I only see three possible states here:

- initial
- ocrOnly
- finished

How can it reach `cardOnly` or `ocrAndCard`?

awushensky: Never mind, I was looking at liveness.

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
kingst: Yup, done.

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,58 @@
...
+//  Created by Sam King on 8/5/20.
+//
+
+import Foundation
+
+public enum MainLoopState {
+    case initial
+    case ocrOnly
+    case cardOnly
+    case ocrAndCard
```
awushensky, line 14: These pertain to liveness only, right? Do we need them for cardscan?

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
kingst: yep, done

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
kingst: right, we'll use these for verify and liveness

awushensky: ahh, so you're planning to share this `MainLoopState` between the three? Would it make more sense to have an individual version of `MainLoopState` for each? It's confusing to have states that can never be reached.

kingst: we could but you can't inherit from enums so the cleanest way would be to make it a class and inherit but I'm open to other suggestions since I do agree with you in principle (but I'm also ok with having it the way that it is currently).

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,59 @@
...
+        let currentState = state
+        state = transition(prediction: prediction)
+        if state != currentState {
+            startTimeForCurrentState = Date()
+        }
+        
+        return state
+    }
+    
+    open func transition(prediction: CreditCardOcrPrediction) -> MainLoopState {
```
jaimejiyepark, line 45: why is this function open? is it going to be overridden?

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
awushensky: in Android, I specifically created separate implementations of a state machine for each product. I started with inheritance, but found that it didn't bring a lot of benefit and added a lot of complexity in implementation.

This is a relatively minor concern, so I don't consider it blocking, but I would personally create a separate state enum and main loop for each product.

jaimejiyepark:
so to my understanding, every time there is a prediction then it checks if the event warrants a transition and if so the state machine class is updated?

CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -0,0 +1,59 @@
...
+//  CardScan
+//
+//  Created by Sam King on 8/5/20.
+//
+
+import Foundation
+
+public enum MainLoopState {
+    case initial
+    case ocrOnly
```
awushensky, line 12: have you considered adding state to these? In the android implementation, I maintained all state, including the voting algorithm, in these states. So when transitioning to an `ocrOnly` state, the state would receive an initial PAN value. When events come in, the state itself keeps track of the current vote.

awushensky: See the CardScan state machine https://github.com/getbouncer/cardscan-android/blob/master/cardscan-ui/src/main/java/com/getbouncer/cardscan/ui/result/MainLoopAggregator.kt#L79

and the liveness state machine https://github.com/getbouncer/cardverify-android/blob/master/liveness-ui/src/main/java/com/getbouncer/liveness/ui/result/MainLoopAggregator.kt#L195

awushensky: This allows the state to be the single `truth` of the system. there's no need to correlate these state enums with the values stored for voting.

CardScan/Classes/CreditCardOcr/CreditCardOcrResult.swift
```diff
@@ -35,7 +37,7 @@ public class CreditCardOcrResult: MachineLearningResult {
     }
     
     public static func finishedWithNonNumberSideCard(prediction: CreditCardOcrPrediction, duration: Double, frames: Int) -> CreditCardOcrResult {
-        let result = CreditCardOcrResult(mostRecentPrediction: prediction, number: "", expiry: nil, name: nil, isFinished: true, duration: duration, frames: frames)
+        let result = CreditCardOcrResult(mostRecentPrediction: prediction, number: "", expiry: nil, name: nil, state: .finished, duration: duration, frames: frames)
```
awushensky, line 40: While it doesn't directly affect this PR, I would advise against having a single `result` class that can have conflicting values. For example, what would it mean to downstream code if it gets a `CreditCardOcrResult` with a valid number and expiry, but is in state `initial`?

It might make sense to have the `state` contain all the possible valid values for that state. For example, `ocrOnly` would have a method to get the current most likely pan, expiry, etc. while `finished` would have a method to get the final pan, expiry, etc. Initial would have no pan nor expiry.

kingst: yeah this function is a huge hack -- let's clean it up later

PR #199, opened 2020-08-06
ActiveStateComputation: make it easy to run ML in active state

kingst:
Our only crashing bug on iOS was due to backgrounding logic and ML. This logic has continued to haunt us and has been the subject of a number of related bugs that we caught before production.

This PR introduces a new class `ActiveStateComputation` that behaves similarly to a serial `DispatchQueue` but runs the computation iff the app is in the active state.

We can use this to protect our main loop ML, warmUp routines, and to enable running completion loop ML with hardware acceleration safely (currently we run completion loop ML on the CPU for safety).


CardScan/Classes/ActiveStateComputation.swift
```diff
@@ -0,0 +1,80 @@
...
+            NotificationCenter.default.addObserver(self, selector: #selector(self.willResignActive), name: UIApplication.willResignActiveNotification, object: nil)
+            NotificationCenter.default.addObserver(self, selector: #selector(self.didBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
+        }
+    }
+    
+    public func async(execute work: @escaping () -> Void) {
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else { return }
+            let state = UIApplication.shared.applicationState
+            guard state == .active, self.isActive else {
```
awushensky, line 56: nit: this should probably not be a guard statement. It would be more readable as a simple if/else.

CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
```diff
@@ -138,16 +129,7 @@ open class OcrMainLoop : MachineLearningLoop {
...
-        let image = UIImage.grayImage(size: CGSize(width: 600, height: 600))
-        let roiRectangle = CGRect(x: 0, y: 0, width: 600, height: 600)
-        guard let cgImage = image?.cgImage else { return }
-        for ocr in mainLoop.analyzerQueue {
-            ocr.dispatchQueue.async {
-                let _ = mainLoop
-                let _ = ocr.recognizeCard(in: cgImage, roiRectangle: roiRectangle)
-            }
-        }
+        // TODO(stk): Implement this later
```
awushensky, line 132: Sanity check: Should this be resolved before meeting this PR?

jaimejiyepark:
makes sense to me. has all those triggers that we thought of not work against this? #lgtm

CardScan/Classes/ActiveStateComputation.swift
kingst: in iOS guard statements have some nice properties, specifically the require you to return after handing the error condition, so we tend to use them a lot (as do the big apps that I have experience with).

CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
kingst: No, this is all going to change a lot after Jaime finishes up dynamic model downloading, so we'll defer until then.

PR #208, opened 2020-08-17
Download and compile model function fix

jaimejiyepark:
the downloaded task's temporary location gets emptied when the completion handler completes so the ml model compiling has to be done in the same function.

### DIFF
--- a/CardScan/Classes/API.swift
+++ b/CardScan/Classes/API.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import CoreML
 
 public struct Api {
     
@@ -135,7 +136,8 @@ public struct Api {
         apiCall(endpoint: endpoint, parameters: apiParameters, completion: completion)
     }
     
-    static public func donwloadLatestModel(signedUrl: String, completion: @escaping ApiCompletion) {
+    @available(iOS 11.0, *)
+    static public func donwloadAndCompileLatestModel(signedUrl: String, completion: @escaping ApiCompletion) {
         guard let url = URL(string: signedUrl) else {
             DispatchQueue.main.async { completion(nil, apiUrlNotSet) }
             return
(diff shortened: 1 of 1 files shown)


CardScan/Classes/API.swift
```diff
@@ -135,20 +136,22 @@ public struct Api {
...
         session.downloadTask(with: url) { (location: URL?, response: URLResponse?, error: Error?) in
-            guard let location = location else {
+            guard let location = location, let compiledUrl = try? MLModel.compileModel(at: location) else {
                 DispatchQueue.main.async { completion(nil, defaultError) }
                 return
             }
+            
             DispatchQueue.main.async {
-                completion(["mlmodel_url": location], nil)
+                completion(["compiled_model_url": compiledUrl], nil)
```
awushensky, line 154: will changing this break any existing code? I couldn't find anything referencing this string, is it used anywhere?

jaimejiyepark: this function is used in model updates in cardverify

awushensky: Got it, so there will be a corresponding change in CardVerify [here](https://github.com/getbouncer/cardverify-ios/blob/7d0918d7681ac66f2d62476a26c0d84b457a2a38/CardVerify/Classes/DynamicModelController.swift#L86)? makes sense, although this is breaking. Do we have any way of indicating which versions of CardVerify are compatible with which versions of CardScan?

jaimejiyepark: yes that is correct. We indicate compatibility in the cardverify podspec file, where cardverify's version dependency of cardscan is stated.

jaimejiyepark:
i can use an `@available` attribute to indicate deprecation but im not sure if this change is it worth it since its only been used internally. but im not sure, what do you think @awushensky?

awushensky:
@jaimejiyepark In this case, let's skip it. For future changes that introduce new APIs, let's mark them as `@available` to indicate that they are experimental features that could change.

PR #210, opened 2020-08-19
Use loop state to drive UI

awushensky:


CardScan/Classes/CreditCardOcr/OcrMainLoop.swift
```diff
@@ -228,7 +230,16 @@ open class OcrMainLoop : MachineLearningLoop {
...
-            delegate?.showCardDetails(number: result.number, expiry: result.expiry, name: result.name)
+            switch (result.state) {
+            case MainLoopState.initial, MainLoopState.cardOnly:
+                delegate?.showNoCard()
+            case MainLoopState.ocrIncorrect:
+                delegate?.showWrongCard(number: result.number, expiry: result.expiry, name: result.name)
+            case MainLoopState.ocrOnly, MainLoopState.ocrAndCard, MainLoopState.ocrDelayForCard:
+                delegate?.showCardDetails(number: result.number, expiry: result.expiry, name: result.name)
+            case MainLoopState.finished:
+                delegate?.complete(creditCardOcrResult: result)
```
kingst, line 241: Does this call `complete` twice?

awushensky: good catch. I'll remove the second call.

PR #213, opened 2020-08-19
Do not capture wrong card

awushensky:


awushensky:
will add a test in a follow-up PR

PR #216, opened 2020-08-27
Add card type extraction

awushensky:


CardScan.podspec
```diff
@@ -19,13 +19,13 @@ CardScan is a library for scanning credit cards.
 
   s.subspec 'Core' do |core|
     core.source_files = 'CardScan/Classes/**/*'
-    core.resource_bundles = { 'CardScan' => ['CardScan/Assets/*.xcassets', 'CardScan/Assets/*.storyboard', 'CardScan/Assets/*.mlmodelc'] }
+    core.resource_bundles = { 'CardScan' => ['CardScan/Assets/*.xcassets', 'CardScan/Assets/*.storyboard', 'CardScan/Assets/*.mlmodelc', 'CardScan/Assets/Config'] }
```
kingst, line 22: I'm surprised that this file pattern works, but the podspec documentation isn't very clear. You could change it to `CardScan/Assets/Config/*.txt` to be consistent but if it works this is fine also.

CardScan/Classes/CreditCardUtils.swift
```diff
@@ -262,6 +264,57 @@ public struct CreditCardUtils {
...
+    public static func determineCardType(cardNumber: String) -> CardType {
+        guard let iin = Int(cardNumber.prefix(6)) else {
+            return .UNKNOWN
+        }
+        
+        let cardTypes: [(ClosedRange<Int>, CardType)]
+        if let cardTypeMap = self.cardTypeMap {
+            cardTypes = cardTypeMap
+        } else {
+            guard let filePath = Bundle.allBundles.compactMap({ $0.path(forResource: "card_types", ofType: "txt", inDirectory: "Config") }).first else {
```
kingst, line 281: We usually use our resource bundle with something that looks like this:
```
guard let bundle = CSBundle.bundle() else { return }
let url = bundle.url(forResource: forResource, withExtension: withExtension)
```

But I'm open to this if you have a reason for using it instead

CardScan.podspec
awushensky: I thought about that, but I didn't want to limit this to just *.txt in case we add different configurations in the future.

CardScan/Classes/CreditCardUtils.swift
awushensky: This is the code I was struggling with for a while. You've just described exactly what I was missing, and I'll use that!

PR #224, opened 2020-09-24
Adding build rules to exclude x86 in iPhone Sim for XCode 12

jaimejiyepark:
To Test:
* build carthage with both XCode 11 and 12 without error
* run example app on simulator and real device without error

I tested with:
* XCode 12: iPhone with iOS14 and sim
* XCode 11.1: iPhone with iOS 13.2 and sim


awushensky (approved):
Looks good for both a real device and a simulator for xcode 12.

PR #231, opened 2020-10-31
Collect flash frames

awushensky:


kingst (approved):
Looks great, ship it!

For cases where the flash is already on does it behave like the current state machine? This might be a question for the verify side.

PR #233, opened 2020-11-09
verif-y @objc wrappers and .xcframework 

jaimejiyepark:


Example/CardScan/ViewController.swift
```diff
@@ -208,13 +216,12 @@ extension ViewController: SimpleScanDelegate {
...
-                            expiryBoundingBox: CGRect?, croppedCardSize: CGSize, squareCardImage:
-                                CGImage, fullCardImage: CGImage, centeredCardState: CenteredCardState?, uxFrameConfidenceValues: UxFrameConfidenceValues?, flashForcedOn: Bool) {}
+extension ViewController: ScanEventsDelegate {
+    func onNumberRecognized(number: String, expiry: Expiry, numberBoundingBox: CGRect, expiryBoundingBox: CGRect, croppedCardSize: CGSize, squareCardImage: CGImage, fullCardImage: CGImage) {}
+    
     func onScanComplete(scanStats: ScanStats) {}
-    func onFrameDetected(croppedCardSize: CGSize, squareCardImage: CGImage, fullCardImage: CGImage,
-                         centeredCardState: CenteredCardState?, uxFrameConfidenceValues: UxFrameConfidenceValues?, flashForcedOn: Bool) {}
+    
+    func onFrameDetected(croppedCardSize: CGSize, squareCardImage: CGImage, fullCardImage: CGImage) {}
```
awushensky, line 224: We need these parameters for verify. I don't think we can safely remove them here.

jaimejiyepark: oh i just replaced that in the example to see if the new delegate worked. With `ScanEventsDelegate` (the objc protocol I made for verif-y), I removed all the `cardverify` related parameters since they're just using `cardscan`

awushensky: ahh I see, I misread this. thanks.

awushensky (approved):

kingst (commented):
The XCFramework is a binary -- I don't think it should go in our source repo. It's ok to have a script for building it in our repo, but not the binary itself. Happy to be convinced otherwise, but generally people don't put binaries in source repos.

awushensky (changes requested):
Aha, that was unclear to me that this was a binary.

CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -392,15 +393,17 @@ public protocol TestingImageDataSource: AnyObject {
         
         let cardSize = CGSize(width: prediction.image.width, height: prediction.image.height)
         if let number = prediction.number, let numberBox = prediction.numberBox {
-            let expiry = prediction.expiryObject()
-            let expiryBox = prediction.expiryBox
+            let expiry = prediction.expiryObject() ?? Expiry(string: "", month: 0, year: 0)
```
kingst, line 396: What is this for? Is it because we can't handle optionals? This feels like a bit of a red flag to me

kingst (commented):
It might be worthwhile for me, you, and Adam to have a chat to figure out what we want to do in the main repo to support this.

jaimejiyepark:
Let me set up a meeting for this since I have more changes to make for this branch for one of our clients and I don't want this PR to just sit

awushensky (approved):

PR #249, opened 2020-12-23
Update system tests to run on simulator and in github workflow

kingst:


jaimejiyepark (commented):
is it possible to sublclass scanbase and add the fake loops instead of modifying it?

kingst:
I couldn't find a way to do that. This loop is protected with both a `testDataSource` and `isSimulator` check so it should be safe, but please do double check that because I agree with your concern

jaimejiyepark:
Tested the system tests out with the simulator and they work. Just to be safe, should we add the `internal` attribute to those functions since they should only be used internally?

jaimejiyepark:
i'll test it once more with the changes

jaimejiyepark (approved):

PR #250, opened 2020-12-23
iPhone 12 CoreML Fix

dxaen:
This PR address the iPhone12 CoreML issue on loading the OCR model. The issue arose from limited support for splicing Tensors on the GPU and/or Neural Engine. Here are the inference times for various devices with this new model.

Device | Inference time + Post processing
------------ | -------------
iPhone 5s | 847ms
iPhone SE | 139ms
iPad Pro 12.9 Second Gen | 50ms
iPad Pro 11 First Gen | 31ms
iPhone 12 mini | 27ms


kingst:
How do these inference times and accuracy compare with the old model?

dxaen:
The inference times are the same but the overall recall dropped from 55.2 to 55.0. Although, I think the videos have changed a bit from the last version.

dxaen:
Interestingly enough, there is drop only on iPhone 12 due to an additional incorrect prediction. For iPhone SE the overall recall and precision for both master and this branch is the same.

jaimejiyepark (approved):

PR #252, opened 2020-12-30
Add translations for the strings in SimpleScanViewController

kingst:
Note: there are a few extra translations in the resource file that we'll use for verify.


jaimejiyepark:
holy moly is this how we support localization? just text files of each translation?

jaimejiyepark (approved):

awushensky (approved):

PR #253, opened 2020-12-31
Add device object

jaimejiyepark:
To match the form event payload on Android, I made a new object that contains just the device info so it can be used in cardverify. Please take a look at the device field payload under `(Client Forms)` https://docs.google.com/document/d/1HvO6VfeafosoWLX9VJggl9mvTLc2-sQuHNeCLTy-lL8/edit#. 

```swift
{
  "locale" : "en-US",
  "ids" : {
    "vendor_id" : "4EC3B517..."
  },
  "phone_count" : 1,
  "platform" : "ios",
  "os_version" : "14.2.1",
  "carrier" : "Verizon",
  "boot_count" : null,
  "phone_type" : 1,
  "name" : "iPhone13,3",
  "network_operator" : null
}
```

#### Todo:
* Do the same with app info
* Unit tests


kingst (commented):
Is it possible to reuse what we already have in the `cardverify-ios` repo? I believe that we have most / all of this, but I could be mistaken. I would also assume that we're going to use the same basic format for this API call that we do for the others?

awushensky (approved):

jaimejiyepark:
we have the various info scattered between cardscan and cardverify. i wanted to unify it into 1 place and i chose to do it in cardscan since we're going to eventually migrate to the v2 payload and i

PR #255, opened 2021-01-12
Customized UI controls

Akoushal:
- Removed skip option
- Removed skip button
- Customized navigation
- Removed back button
- Customized podspec


Akoushal:
Created by mistake

kingst:
@Akoushal you're using an old version of CardScan as the base for your project, I highly recommend switching to a newer version -- our performance is much much better now.

PR #258, opened 2021-01-28
[RFC] Add fast/accurate flag to scan config

jaimejiyepark:


awushensky (approved):

CardScan/CardScan/Classes/ScanConfiguration.swift
```diff
@@ -1,6 +1,12 @@
 import Foundation
 
+@objc public enum ScanPerformance: Int {
+    case fastScan
```
kingst, line 4: Nit: I'd call these `fast` and `accurate` -- drop the Scan

CardScan/CardScan/Classes/ScanConfiguration.swift
```diff
@@ -1,6 +1,12 @@
...
+    case accurateScan
+}
+
 @objc public class ScanConfiguration: NSObject {
     @objc public var runOnOldDevices = false
     @objc public var setPreviouslyDeniedDevicesAsIncompatible = false
+    @objc public static var scanPerformancePriority: ScanPerformance = .fastScan
```
kingst, line 11: I wouldn't make this a static configuration variable -- it should be a member variable on `SimpleScanViewController` that defaults to `fast` and people can pick `accurate` on a per viewcontroller basis. For example, if someone is using both scan and verify they might want different configurations for the two.

kingst (commented):
I wouldn't set this as a static variable. Also, is there a simple unit test that might make sense for this? Any state machine updates make me nervous.

jaimejiyepark:
@kingst I made changes in that instead of changing `OcrMainLoopStateMachine` , `SimpleScan` will use the new `OcrDurationMainLoopStateMachine` (which is the modified custom state machine we sent to verif-y and gsa)  when set to `.accurate`.

CardScan/CardScan/Classes/CreditCardOcr/MainLoopStateMachine.swift
```diff
@@ -67,3 +66,51 @@ public protocol MainLoopStateMachine {
         return OcrMainLoopStateMachine()
     }
 }
+
+@objc public class OcrDurationMainLoopStateMachine: NSObject, MainLoopStateMachine {
```
kingst, line 70: Nit: Maybe call this `OcrAccurateMainLoopStateMachine` or something like that

CardScan/CardScan/Classes/ScanConfiguration.swift
```diff
@@ -1,5 +1,10 @@
 import Foundation
 
+@objc public enum ScanPerformancePriority: Int {
```
kingst, line 3: Nit: Maybe call this `ScanPerformance`

kingst (approved):
A few optional comments, but ship it.

Can you add another button to the example app to test out `accurate` mode?

After you address these ship it

PR #268, opened 2021-04-01
Added script to update sdk version to info.plist

jaimejiyepark:
I tested by:
1. running `update_version.sh 2.0.15`
2. running `build_framework.sh`
3. running example app with newly created .xcframework
4. checking GCP for stat to come in and check sdk_version to see if its `2.0.15`
Screen Shot 2021-04-01 at 12 03 36 PM

### DIFF
--- a/scripts/deploy_xcframework.sh
+++ b/scripts/deploy_xcframework.sh
@@ -36,6 +36,8 @@ fi
 PROD_BRANCH="production-$(date +"%Y%m%d-%s")"
 git checkout -b $PROD_BRANCH
 
+./scripts/update_version.sh ${1}
+
 ./scripts/build_xcframework.sh
 
 ./scripts/setup_package.sh ${PROD_BRANCH} ${PROD_BRANCH}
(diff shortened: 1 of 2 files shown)


scripts/deploy_xcframework.sh
```diff
@@ -36,6 +36,8 @@ fi
 PROD_BRANCH="production-$(date +"%Y%m%d-%s")"
 git checkout -b $PROD_BRANCH
 
+./scripts/update_version.sh ${1}
```
awushensky, line 39: do we need to create a new commit before cutting the branch?

jaimejiyepark: wouldn't the changes be added in line 61?

awushensky: ahh, yes. they would. it won't go to the master branch, but will be pushed to the prod branch. that's probably fine.

awushensky (approved):

scripts/deploy_xcframework.sh
jaimejiyepark: would be it be better practice to push it to master?

PR #273, opened 2021-04-27
Update scan stats with new API request class

jaimejiyepark:


awushensky (approved):

CardScan/CardScan/Classes/ScanStats.swift
```diff
@@ -63,6 +63,19 @@ public struct ScanStats {
...
+    func createPayload() -> ScanStatisticsPayload {
+        let scanStats = ScanStatistics(scans: self.scans,
+                                       cardsDetected: self.cardsDetected,
+                                       torchOn: self.torchOn,
+                                       orientation: self.orientation,
+                                       success: self.success,
+                                       duration: self.duration(),
+                                       model: self.model,
+                                       permissionGranted: self.permissionGranted,
+                                       userCanceled: self.userCanceled)
```
awushensky, line 75: maybe this is iOS style, but I really don't like hanging indents like this. It's not guaranteed to be a multiple of four and can break easily by changing anything in the first line. instead, I greatly prefer the style:
```
let scanStats = ScanStatistics(
    scans: self.scans,
    ...
    userCanceled: self.userCanceled,
)
```

jaimejiyepark: I think it depends on the style guide of each place. I just looked different companies guides on multi-line styling and a lot of them just use the xcode indentation. But it's not that big of a deal right now so I can change it to how you suggest it :-)

CardScan/CardScan/Classes/ScanBaseViewController.swift
```diff
@@ -402,7 +408,13 @@ public protocol TestingImageDataSource: AnyObject {
         if (self.sendScanStats) {
             // fire and forget
             let scanStatsPayload = self.ocrMainLoop()?.scanStats.createPayload() ?? ScanStats().createPayload()
-            ScanApi.uploadScanStats(payload: scanStatsPayload, completion: { _, _ in })
+            ScanApi.uploadScanStats(payload: scanStatsPayload, completion: { response, error in
+                guard let status = response?.status, status == "ok" else {
```
awushensky, line 412: is this guaranteed to always be lowercase?

awushensky (approved):

jaimejiyepark:
The integration test relies on this line https://github.com/getbouncer/cardscan-ios/blob/f37f24b041906560242eeb466911e29f5d8da20f/CardScan/CardScan/Classes/API.swift#L78. 

I don't want to change the integration tests right now and in this PR so I added the `lastScanStatsSuccess` value in the codebase but I plan to remove it and update the tests

CardScan/CardScan/Classes/ScanBaseViewController.swift
jaimejiyepark: yea it is https://github.com/getbouncer/cardscan-ios/blob/f37f24b041906560242eeb466911e29f5d8da20f/CardScan/CardScan/Classes/API.swift#L77-L78

PR #278, opened 2021-05-05
Fix failing unit tests

jaimejiyepark:


.github/workflows/unit_test.yml
```diff
@@ -25,17 +25,6 @@ jobs:
...
-        run:
-          xcodebuild test -workspace CardScan.xcworkspace -scheme CardScanTests -destination 'platform=iOS Simulator,name=iPhone 11'
-
-      - name: upload-artifacts-unit
-        uses: actions/upload-artifact@v2
-        if: failure()
-        with:
-          name: test-report-unit
-          path: /Users/runner/Library/Developer/Xcode/DerivedData/*/Logs/Test/*.xcresult
-
```
awushensky, line 38: this will prevent unit tests from running. is this what we want?

jaimejiyepark: All the unit tests (including the new payload test) are in `CardScanExampleTests`. Its okay to remove the `CardScanTests` scheme!

jaimejiyepark: The new unit test among the older tests
Screen Shot 2021-05-05 at 3 50 02 PM

awushensky (approved):

PR #287, opened 2021-06-17
Adds calculation to infer boxes for each digit with apple OCR

kingst-stripe:
AppleOCR gives us only one bounding box for the full PAN, this PR assumes that the font is fixed width and infers each digit bounding box. The AppleOCR bounding boxes aren't great in general if the card is rotated, but for generally straight cards this works reasonably well.

### DIFF
--- a/CardScan/CardScan/Classes/CreditCardOcr/AppleCreditCardOcr.swift
+++ b/CardScan/CardScan/Classes/CreditCardOcr/AppleCreditCardOcr.swift
@@ -21,6 +21,7 @@ public class AppleCreditCardOcr: CreditCardOcrImplementation {
         let startTime = Date()
         var name: String?
         var nameBox: CGRect?
+        var numberBoxes: [CGRect]?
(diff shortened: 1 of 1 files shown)


jaimepark-stripe:
Ahha so a 15 digit card is [1234][s][567890][s][12345] = length 17 and a 16 digit card is [1234][s][5678][s][9012][s][3456]  = length 19. You can get every digit by the height of the bigger bounding box and width of every digit by the length of the total amount of characters. 

What will this be used for?

kingst-stripe:
Good question. This interface was never correct for AppleOcr so this PR fixes that bug, but the reason that this came up is that we'll use this functionality for middle six redaction.

jaimepark-stripe:
Is there a reason why we're using Apple OCR instead of our OCR?

kingst-stripe:
We use both, but before this PR the logic for Apple didn't work

jaimejiyepark (approved):

